PR #3, opened 2018-09-10
Submition to challenge

talvasconcelos:
Just leaving my submission to the challenge. I didn't have the time (current job got in the way) to finished all up. I'll send an email explaining it better.

Best regards,
Tiago


my-code/movie/src/App.js
```diff
@@ -0,0 +1,74 @@
...
+      detail: "",
+      fav: [],
+      results: movies.Search//false
+    }
+
+    this.changeText = this.changeText.bind(this)
+    this.getMovies = this.getMovies.bind(this)
+    this.getDetails = this.getDetails.bind(this)
+    this.handleLikes = this.handleLikes.bind(this)
+  }
```
danilowoz, line 26: You don't need to bind every function in the `constructor` anymore. You can remove all bind functions and even the state, then set the function as arrow function, thanks to Public Class Field. 

So your app will look something like this:
```jsx
class App extends Component {
  state = { }

  handleLikes () => { 
    this.setState({})
  }

  render() { /* */ }
}
```

Writing class this way, you won't repeat too much code.
[Read more](https://medium.com/@nikolalsvk/loosing-bind-this-in-react-8637ebf372cf)

my-code/movie/src/App.js
```diff
@@ -0,0 +1,74 @@
...
+
+  handleLikes(id) {    
+    const remove = (filmID) => {this.state.fav.filter(fav => {
+      return fav !== filmID
+    })}
+
+    if(this.state.fav.includes(id)){
+      let newFav = remove(id)
```
danilowoz, line 34: Prefer to use `const` instead of` let`. [Read more](https://hackernoon.com/js-var-let-or-const-67e51dbb716f) about the differences between them

my-code/movie/src/App.js
```diff
@@ -0,0 +1,74 @@
...
+      return this.setState({fav: newFav})
+    }
+    this.setState({fav: [id, ...this.state.fav]})
```
danilowoz, line 37: You used the `fav` state three times here, you might create a `const` in the top of the function:

```jsx
const { fav } = this.state
```

Again, your code will look shorter and with better performance.

my-code/movie/src/App.js
```diff
@@ -0,0 +1,74 @@
...
+          <Route exact path = '/' render={(props) => <Home {...props} res={this.state.results} search={this.state.value} searchHandler={this.changeText} submit={this.getMovies}/>} />
+          <Route path = '/details/:id' render={(props) => <Details {...props} fetch={this.getDetails} detail={this.state.detail} like={this.handleLikes}/>} />
+        </div>
+      </Router>
+      
+    );
+  }
+}
+
+export default App
```
danilowoz, line 74: I think you have to choose a way to manage your data. Currently, there are a lot of options, such as Redux, React Context, MobX, etc. At here, in Significa, we prefer to use Redux because we can control the side effect, it has a single data store, it's immutable, plus there are too many tools that help our work. 

So, think about these options. It will help your job a lot  :)
Take a look at EggHead and Udemy, there are some amazing courses there!

my-code/movie/src/api.js
```diff
@@ -0,0 +1,13 @@
+// http://www.omdbapi.com/?i=tt3896198&apikey=5c129d0b
+const API_KEY = '&apikey=5c129d0b' //should be on an .env
+const URL = 'http://www.omdbapi.com/?'
+
+export const movieSearch = (search) => {
+    return fetch(`${URL}s=${search}${API_KEY}`)
+        .then(res => res.json())
```
danilowoz, line 7: Don't forget to catch the errors

my-code/movie/src/api.js
```diff
@@ -0,0 +1,13 @@
+// http://www.omdbapi.com/?i=tt3896198&apikey=5c129d0b
+const API_KEY = '&apikey=5c129d0b' //should be on an .env
+const URL = 'http://www.omdbapi.com/?'
+
+export const movieSearch = (search) => {
+    return fetch(`${URL}s=${search}${API_KEY}`)
+        .then(res => res.json())  
+}
+
+export const movieDetails = (id) => {
```
danilowoz, line 10: Same here

my-code/movie/src/components/like.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+
+export const Faved = () => {
```
danilowoz, line 3: Minify the SVG, there are a lot of unnecessary code

my-code/movie/src/index.css
```diff
@@ -0,0 +1,244 @@
+:root {
+  /* Colors */
+  --yellow: #ff9f1c;
+  --red: #ff4040;
+  --green: #2ec4b6;
+  --dark: #0a1014;
+  --grey: #1b2329;
+  --mid-grey: #353f4c;
+  --light-grey: #7a8c99;
+  --white: #fff;
```
danilowoz, line 10: Nice CSS vars 👍

my-code/movie/src/index.css
```diff
@@ -0,0 +1,244 @@
...
+}
+
+@import url('https://fonts.googleapis.com/css?family=Roboto:400,500,700');
```
danilowoz, line 13: Avoid use imports in CSS, because it'll block that the next CSS file is loaded

my-code/movie/src/index.css
```diff
@@ -0,0 +1,244 @@
...
+
+html, body {
+	margin: 0;
+	padding: 0;
+	box-sizing: border-box;
+	height: 100%;
+	font: calc(100% + 0.1vw)/1.5 Roboto, -apple-system, system-ui, BlinkMacSystemFont, "Segoe UI";
```
danilowoz, line 20: Cool! What exactly `calc(100% + 0.1vw)` means?

my-code/movie/src/index.css
```diff
@@ -0,0 +1,244 @@
...
+.btn {
+  display: inline-block;
+  text-align: center;
+  white-space: nowrap;
+  vertical-align: middle;
+  box-sizing: border-box;
+  border: 1px solid var(--mid-grey);
+  border-radius: 5px;
+  font-size: 1rem;
+  padding: 0.5em 0.75em;
```
danilowoz, line 49: Don't mess up the units, choose only one type.
I like a lot the idea about `em` and `rem`, and I used to set for `font-size` the `rem` (root/body `em`) and paddings and margins as `em`.

my-code/movie/src/index.css
```diff
@@ -0,0 +1,244 @@
...
+  opacity: 0;
+  height: 100%;
+  margin: 0;
+  padding: 12px;
+  color: var(--white);
+}
+
+.info:hover {
+  opacity: 1;
+  transition: 300ms ease;
```
danilowoz, line 118: For opacity transitions, the `ease-in-out` looks better

danilowoz (commented):
Thanks, @talvasconcelos for your submission. 🍻 

I left some comments that I think you could improve in your application, most of them are just details. But take a look about the management data store (Redux, etc) and CSS methodologies (BEM, ITCSS, OOCSS) or even about the CSS-in-js, I think you'll enjoy learning more about it :)

my-code/movie/src/App.js
talvasconcelos: Didn't knew about that, i'll have a look at it.

my-code/movie/src/App.js
talvasconcelos: I see the point, and tought about it. Sometimes i get carried away about getting some results and try to rush it without thinking. After i get it done and working i tend to come back and clean the code.

my-code/movie/src/App.js
talvasconcelos: I know about Redux, never used it before because i never felt the need. I build small apps and i guess Redux could be an overkill. I've been wanting to dive in for Redux for quite some time...

my-code/movie/src/components/like.js
talvasconcelos: Always have a hard time around SVG...

my-code/movie/src/index.css
talvasconcelos: The idea is that the font size becomes responsive. Inittially i had calc(100% + 0.25vw) but it was getting the font size too big. 100% beeing normally 16px, it grows when the screen gets wider. There's a way to truncate to a maximum with some more math also.

my-code/movie/src/index.css
talvasconcelos: That was probabilly a typo and was suposed to be rem

talvasconcelos:
As soon as i can i'll review the code taking into account your recommendations . Thanks for taking the time to look at the messy code. I rushed it out because i'm currently with some workload on my day job and would like to send the application to the challenge.

Best regards,
Tiago

danilowoz:
I'm going to close this conversation, feel free to submit a new PR again or reopen it whenever you want

Thanks for your submission 🍻

PR #4, opened 2018-09-12
Robbie Hilado - Front End Challenge

sopasdigital:
This is my attempt at doing the front end challenge for Signfica.

I was able to create the core functionality of the app except for 'favourite' functionality. Given the time limit, I was not able to apply all the proper styling presented in the style guide. I'm also aware that I'm missing some core states for the components like when there's no input from the user as of yet, if there are no poster images available to display, etc.

Danilo was a big help as he gave me some tips and suggestions on how to go about sending necessary state props to display information on the movie details page.


pdrbrnd:
Thanks for the submission Robbie 🎉

my-code/movie-search/src/Actions/index.js
```diff
@@ -0,0 +1,25 @@
+import axios from 'axios';
+
+export const FETCH_MOVIES = 'FETCH_MOVIES';
+export const FETCH_MOVIE = 'FETCH_MOVIE';
+
+const ROOT_URL = 'http://www.omdbapi.com/?apikey=';
+const API_KEY = '5766cd6e';
+
+export function fetchMovies(searchTerm) {
+  const request = axios.get(`${ROOT_URL}${API_KEY}&s=${searchTerm}`);
```
pdrbrnd: given that `axios` returns a Promise, you're potentially giving your reducer a pending Promise payload.

The goal for these kind of features is to make sure you can take care of asynchronous calls (like an API request). Async calls have two distinct moments:
1. When you start to make the call
2. When there's an answer (or timeout).

It's always a good practice to handle correctly and create appropriate UI for these two moments by, for instance, creating an `isFetching` flag in your reducers or component state to show a loader when needed.
Also, we need to catch any kind of information if the request fails for some reason.

A simple solution (albeit very naïve, untested and whatnot) would be something like:
```jsx
class YourComponent extends Component {
  state = { 
    searchTerm: '',
    isFetching: false,
    results: [],
    error: false,
  }

  fetchMovies = () => {
    this.setState({ isFetching: true })

    axios.get(`${ROOT_URL}${API_KEY}&s=${searchTerm}`)
    .then(response => this.setState({ results: response.data, isFetching: false }))
    .catch(() => this.setState({ error: true, isFetching: false }))
  }

  // ...
}
```

If Redux is your thing, `redux-thunk` is a simple middleware that allows you to return functions for your action creators, thus making possible the delay of the dispatch.

```jsx
export const fetchMovies = searchTerm => dispatch => {
  dispatch({ type: 'FETCH_MOVIES_START' });

  axios.get(`${ROOT_URL}${API_KEY}&s=${searchTerm}`)
  .then(response => dispatch({ type: 'FETCH_MOVIES_COMPLETE', payload: response.data }))
  .catch(error => dispatch({ type: 'FETCH_MOVIES_ERROR', payload: error, error: true }))
}
```
or something like this.
This way we have 3 distinct action types (start, complete, error) and can make the UI react to each situation

(All this code was written here in the comments and is naive and untested)

my-code/movie-search/src/App.js
```diff
@@ -0,0 +1,33 @@
+import React, { Component } from 'react';
+import { Provider } from 'react-redux';
+import { createStore, applyMiddleware } from 'redux';
+import promise from 'redux-promise';
+import { BrowserRouter, Route, Switch } from 'react-router-dom';
+
+import reducers from './Reducers';
```
pdrbrnd: The Capitalized words have a semantic meaning and I would avoid naming folders like "Actions", "Reducers", "Components" or "Containers".

It's best to leave capitalisation to components only

my-code/movie-search/src/App.js
```diff
@@ -0,0 +1,33 @@
...
+import AppLogo from './Components/AppLogo';
+import LandingPage from './Layouts/LandingPage';
+import MovieDetails from './Containers/MovieDetails';
+import './App.css';
+
+const createStoreWithMiddleware = applyMiddleware(promise)(createStore);
+
+class App extends Component {
```
pdrbrnd: this can be refactored as a functional component

my-code/movie-search/src/Assets/icons/icon-arrow-grey.svg
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
pdrbrnd: A bit off-topic but it's always a good idea to compress svgs using something like SVGOMG

my-code/movie-search/src/Components/AppLogo/index.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+const AppLogo = (props) => <div className="AppLogo"><img src={require('../../Assets/logos/logo.svg')} alt="What's in"/></div>
```
pdrbrnd: Why this `require`?

You could
```jsx
import React from 'react';
import logo from '../../Assets/logos/logo.svg';
```

my-code/movie-search/src/Containers/MovieCatalogue/index.js
```diff
@@ -0,0 +1,42 @@
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import _ from 'lodash';
```
pdrbrnd: You are importing a very big library to use just one function (`map`).
Besides, map is a ES6 function that you can use without lodash.

Honestly, most of lodash's functions can be easily dropped in favour of ES6

my-code/movie-search/src/Containers/MovieCatalogue/index.js
```diff
@@ -0,0 +1,42 @@
...
+import { Link } from 'react-router-dom';
+
+import { fetchMovies } from '../../Actions';
```
pdrbrnd: You're importing this action creator but it's never used

my-code/movie-search/src/Containers/MovieCatalogue/index.js
```diff
@@ -0,0 +1,42 @@
...
+
+class MovieCatalogue extends Component {
+  renderCatalogue() {
```
pdrbrnd: There's no need for `renderCatalogue`.
This class could be a functional component with the mapping of the movies directly in the render:

```jsx
const MovieCatalogue = ({ moviesList }) => (
  <div className="MovieCatalogue">
    <ul>
      {moviesList.Search.map(movie => (
        <li key={movie.imdbID} className="Card">
        <Link to={`/movies/${movie.imdbID}`}>
          <div className="Card-Overlay">
            <div className="Card-Info">
              <h4 className="Card-Title">{movie.Title}</h4>
              <span className="Card-Year">{movie.Year}</span>
            </div>
          </div>
          <img className="Card-Poster" src={movie.Poster} alt={movie.Title} />
        </Link>
      </li>
      ))}
    </ul>
  </div>
);
```

my-code/movie-search/src/Containers/MovieDetails/index.js
```diff
@@ -0,0 +1,66 @@
...
+import { Link } from 'react-router-dom';
+
+import { fetchMovie } from '../../Actions';
+
+class MovieDetails extends Component {
+  componentDidMount() {
+    const { imdbID } = this.props.match.params;
+    this.props.fetchMovie(imdbID);
+
+    console.log(this.props);
```
pdrbrnd: Careful with the console.logs and commented code (assuming that this submission is like a PR to production)

my-code/movie-search/src/Containers/MovieDetails/index.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  render() {
+    const movie = this.props.chosenMovie;
```
pdrbrnd: you could use ES6 destructuring: https://codeburst.io/es6-destructuring-the-complete-guide-7f842d08b98f?gi=d93f8df8d0a1

```jsx
const { chosenMovie: movie } = this.props;
```

my-code/movie-search/src/Containers/MovieDetails/index.js
```diff
@@ -0,0 +1,66 @@
...
+    const getValues = movie.Ratings&&movie.Ratings.map((review) => <span key={review.Source} className="Reviews">{review.Source} : {review.Value}</span> );
```
pdrbrnd: This is very long and hard to read. Consider using Prettier or something like it.
Also, using a linter would've helped you in general.

my-code/movie-search/src/Containers/MovieDetails/index.js
```diff
@@ -0,0 +1,66 @@
+import React, { Component } from 'react';
```
pdrbrnd: As you're not dispatching several actions on your async calls, if we open a movie detail, go back and open another movie, we still see the old details for a second before it changes

my-code/movie-search/src/Containers/SearchBar/index.js
```diff
@@ -0,0 +1,34 @@
...
+class SearchBar extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = { searchTerm: '' };
+  }
+
+  onInputChange(searchTerm) {
+    this.setState({ searchTerm });
+    this.props.fetchMovies(searchTerm);
```
pdrbrnd: We're making A LOT of requests here!
consider throttling or debouncing these requests

my-code/movie-search/src/Containers/SearchBar/index.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  render() {
+    return (
+      <div className="SearchBar">
+        <input
+          placeholder="Search movies..."
+          value={this.state.searchTerm}
+          onChange={event => this.onInputChange(event.target.value)} />
```
pdrbrnd: You could just pass the function reference and get event.target.value there instead of creating a new anonymous function on each render

pdrbrnd:
@sopasdigital Left some comments. @ruicsereno will also review your CSS

my-code/movie-search/src/App.css
```diff
@@ -0,0 +1,159 @@
...
+h3 {
+  font-size: 24px;
+}
+
+h4 {
+  font-size: 20px;
+  font-weight: 700;
+}
+
+* {
```
ruicsereno: I'd use this kind of generic rules at a Reset separate file.

my-code/movie-search/src/App.css
```diff
@@ -0,0 +1,159 @@
+@import url('https://fonts.googleapis.com/css?family=Roboto:300,400,500,700');
+
+body {
+  background: #0A1014;
+  font-size: 16px;
+  font-family: 'Roboto', sans-serif;
+  color: #7A8C99;
+}
+
+h1 {
```
ruicsereno: Typography rules could be a separate file for better organisation

my-code/movie-search/src/App.css
```diff
@@ -0,0 +1,159 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  list-style: none;
+  padding: 0;
+  margin: 30px 0;
+  justify-content: center;
+
+}
+
+.Card {
```
ruicsereno: All Card rules could be in a separate file

my-code/movie-search/src/App.css
```diff
@@ -0,0 +1,159 @@
...
+}
+:-moz-placeholder { /* Firefox 4 - 18 */
+  color: #7A8C99;
+  opacity: 1;
+}
+
+.MovieCatalogue ul {
+  display: flex;
+  flex-wrap: wrap;
+  list-style: none;
```
ruicsereno: List style could be solved at the Reset file. It would make CSS rules to the point and cleaner

my-code/movie-search/src/App.css
```diff
@@ -0,0 +1,159 @@
...
+  padding: 0;
```
ruicsereno: Same as list-style

my-code/movie-search/src/Actions/index.js
sopasdigital: This is great information, Pedro! I was wondering what would be good way of catching these eventualities. I'd love to add this to what I currently have.

I have yet to try `redux-thunk`, but I think it would be a cool inclusion. I still need to grasp the idea of states, I guess.

Would it be better I learn the 'simple solution' method first so I understand what's happening in vanillaJS before using `redux-thunk`?

my-code/movie-search/src/App.css
```diff
@@ -0,0 +1,159 @@
...
+  object-fit: cover;
+  height: 330px;
+}
+
+.MovieDetails {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.MovieDetails-Information {
```
ruicsereno: Each component, not only this ones and the ones I commented above could be separate, at different files. It would be easier to locate them and to work on it specifically.

my-code/movie-search/src/App.js
sopasdigital: Ah! Okay, duly noted!

my-code/movie-search/src/App.css
```diff
@@ -0,0 +1,159 @@
...
+  width: 60%;
+}
+
+.MovieDetails-Information p {
```
ruicsereno: This shouldn't be done. Typography should all be defined at a Typography file. Styling text like this, so specifically, is not error proof at all. Avoid such specific styling

ruicsereno (changes requested):

my-code/movie-search/src/Assets/icons/icon-arrow-grey.svg
sopasdigital: Ohh! I'll look into that, thanks for the suggestion!

my-code/movie-search/src/Components/AppLogo/index.js
sopasdigital: You're right, I should've done that instead.

my-code/movie-search/src/Containers/MovieCatalogue/index.js
sopasdigital: Oh, that makes sense. I thought lodash was necessary for this case. I'll refactor this then :)

my-code/movie-search/src/Containers/MovieCatalogue/index.js
sopasdigital: Right, I wasn't able to take it out. Good catch!

my-code/movie-search/src/Containers/MovieCatalogue/index.js
sopasdigital: I'll probably need you, Danilo or whoever to help me grasp the difference on this one. Thanks for pointing it out though!

my-code/movie-search/src/Containers/MovieDetails/index.js
sopasdigital: Yes, I noticed this too. Your earlier recommendation on async catches this eventuality?

my-code/movie-search/src/Containers/MovieDetails/index.js
sopasdigital: Duly noted!

my-code/movie-search/src/Containers/MovieDetails/index.js
sopasdigital: This will be a must-read for me. I've only started understanding when I can destructure.

my-code/movie-search/src/Containers/SearchBar/index.js
sopasdigital: Right. I wasn't able to add the debounce on this one, my bad.

my-code/movie-search/src/Containers/SearchBar/index.js
sopasdigital: Noted!

my-code/movie-search/src/App.css
sopasdigital: I initially wanted to separate each of the css based on which components used them. I'll ask the team how they typically structure these files so I can apply that. Thanks for the feedback, Rui!

my-code/movie-search/src/actions/index.js
```diff
@@ -0,0 +1,43 @@
...
+export const types = {
+  FETCH_CATALOGUE_START: "FETCH_CATALOGUE_START",
+  FETCH_CATALOGUE_COMPLETE: "FETCH_CATALOGUE_COMPLETE",
+  FETCH_CATALOGUE_ERROR: "FETCH_CATALOGUE_ERROR",
+  FETCH_MOVIE_START: "FETCH_MOVIE_START",
+  FETCH_MOVIE_COMPLETE: "FETCH_MOVIE_COMPLETE",
+  FETCH_MOVIE_ERROR: "FETCH_MOVIE_ERROR"
+};
+
+const ROOT_URL = "http://www.omdbapi.com/?apikey=";
```
danilowoz, line 12: You can create an instance from Axios, then you don't need to set the URL every time:
https://github.com/axios/axios#axioscreateconfig

my-code/movie-search/src/actions/index.js
```diff
@@ -0,0 +1,43 @@
...
+const API_KEY = "5766cd6e";
```
danilowoz, line 13: You should never send the private keys to Github, save it in a `.env` file
https://www.npmjs.com/package/dotenv

my-code/movie-search/src/components/MovieCatalogue/index.js
```diff
@@ -0,0 +1,35 @@
...
+                <Text>{movie.Year}</Text>
+              </CardInfo>
+            </CardOverlay>
+          </Link>
+        </Card>
+      ))}
+  </Container>
+);
+
+function mapStateToProps({ movieCatalogue }) {
```
danilowoz, line 28: Prefer ES6 syntax:
```jsx
const mapStateToProps = ({ movieCatalogue }) => ({ movieCatalogue })
```

my-code/movie-search/src/components/UI/Provider/index.js
```diff
@@ -0,0 +1,43 @@
...
+  	-ms-text-size-adjust: 100%;
+  	-webkit-font-smoothing: antialiased;
+    -webkit-overflow-scrolling: touch;
+    -webkit-tap-highwhite-color: transparent;
+  	-webkit-text-size-adjust: 100%;
+  }
+
+  body {
+    background-color: ${colors.greyscale.black};
+    font-family: 'Roboto', sans-serif;
```
danilowoz, line 24: Save the main font family in a variable

my-code/movie-search/src/components/UI/Provider/utils.js
```diff
@@ -0,0 +1,30 @@
+import colors from "./colors";
```
danilowoz, line 1: Split the utils in different files

my-code/movie-search/src/components/UI/Provider/utils.js
```diff
@@ -0,0 +1,30 @@
...
+
+export const applyColor = props => {
```
danilowoz, line 3: I'm not a big fan of `swith case` :)
Maybe you can try something like this:

```js
export default ({color}) => {
  const variations = {
    yellow: colors.yellow,
    red: colors.red,
    ....
  }

  return colors[color] ? colors[color] : colors.greyscale.black
}
```

my-code/movie-search/src/containers/MovieDetails/index.js
```diff
@@ -0,0 +1,113 @@
...
+import RottenTomatoes from "../../components/UI/Logos/RottenTomatoes";
+
+class MovieDetails extends Component {
+  componentDidMount() {
+    const { imdbID } = this.props.match.params;
+
+    this.props.fetchMovie(imdbID);
+  }
+
+  convertToList(list, title) {
```
danilowoz, line 28: Prefer to use arrow function and you should always return something.
In this case, you can early return `null` whether `list` is `null` or `undefined`.

Take a look at this: https://forum.freecodecamp.org/t/the-return-early-pattern/19364

my-code/movie-search/src/containers/MovieDetails/index.js
```diff
@@ -0,0 +1,113 @@
...
+                <Text>{listItem}</Text>
+              </li>
+            ))}
+          </ul>
+        </Container>
+      );
+    }
+  }
+
+  applyIcon(src) {
```
danilowoz, line 47: Same thing here, about the early return and `swith case`

my-code/movie-search/src/containers/MovieDetails/index.js
```diff
@@ -0,0 +1,113 @@
...
+          <Container justifyRight width="40%">
+            <ImageHolder height="530px" width="350px" bg={movie.Poster} />
+          </Container>
+        </Container>
+      </Template>
+    );
+  }
+}
+
+function mapStateToProps(state) {
```
danilowoz, line 106: ES6 syntax please

my-code/movie-search/src/containers/SearchBar/index.js
```diff
@@ -0,0 +1,49 @@
...
+
+import { fetchCatalogue } from "../../actions";
+import Input, { SearchBarContainer, Icon } from "../../components/UI/Inputs";
+
+class SearchBar extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = { searchTerm: "" };
+    this.startFetch = debounce(this.startFetch, 300);
```
danilowoz, line 13: 👍

my-code/movie-search/src/containers/SearchBar/index.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  onInputChange = e => {
```
danilowoz, line 16: maybe: 
```js
onInputChange = ({target: { value: searchTerm }}) => {
    this.setState({ searchTerm }, this.startFetch);
}
```

my-code/movie-search/src/containers/SearchBar/index.js
```diff
@@ -0,0 +1,49 @@
...
+          placeholder="Search movies..."
+          value={searchTerm}
+          onChange={this.onInputChange}
+        />
+      </SearchBarContainer>
+    );
+  }
+}
+
+function mapStateToProps({ movieCatalogue }) {
```
danilowoz, line 42: ES6 syntax please

my-code/movie-search/src/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+import registerServiceWorker from "./registerServiceWorker";
+
+const store = createStore(
+  reducers,
+  window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
+  applyMiddleware(thunk)
+);
+
+// const createStoreWithMiddleware = applyMiddleware(thunk)(createStore);
```
danilowoz, line 21: remove the unnecessary comment

my-code/movie-search/src/index.js
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+import ReactDOM from "react-dom";
+import { Provider as Store } from "react-redux";
```
danilowoz, line 3: Save the redux things in a different file, then you will split the responsibilities of this file

my-code/movie-search/src/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+import reducers from "./reducers";
+import Provider from "./components/UI/Provider";
+import Routes from "./routes";
+
+import registerServiceWorker from "./registerServiceWorker";
+
+const store = createStore(
+  reducers,
+  window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
```
danilowoz, line 17: Only add this line to dev mode, read more about environments mode in Node.js

my-code/movie-search/src/reducers/reducerChosenMovie.js
```diff
@@ -0,0 +1,13 @@
+import { types } from "../actions";
+
```
danilowoz, line 2: You can destructure the variable on the top of the function and please use ES6 syntax

```js
export default (state = {}, { type, payload }) => {
```

my-code/movie-search/src/reducers/reducerMovieCatalogue.js
```diff
@@ -0,0 +1,13 @@
+import { types } from "../actions";
+
+export default function(state = {}, action) {
```
danilowoz, line 3: same thing here

my-code/movie-search/src/routes/index.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { Switch, Route } from "react-router-dom";
+
+import MovieSearch from "../views/MovieSearch";
+import MovieDetails from "../containers/MovieDetails";
+
+export default () => (
```
danilowoz, line 7: Remember to catch the page not found

danilowoz (changes requested):

my-code/movie-search/src/actions/index.js
sopasdigital: Oh nice! I'll take a look at this :)

my-code/movie-search/src/actions/index.js
sopasdigital: ah, noted!

my-code/movie-search/src/components/UI/Provider/utils.js
sopasdigital: will do!

my-code/movie-search/src/routes/index.js
sopasdigital: I'll need your help and guidance for this one :)

my-code/movie-search/src/index.js
sopasdigital: Oh okay, I thought this was necessary for the actual app. Will note that :)

PR #6, opened 2018-10-25
Submission

isabelsa:
Unfinished submission, missing loads.


my-code/src/components/Button/index.js
```diff
@@ -0,0 +1,18 @@
...
+import RegularButton from "./styled"
+import Heart from "../Icons/Heart"
+
+type PropsType = {
+  text: string
+}
+
+const Button = ({ text, ...props }: PropsType) => (
+  <RegularButton {...props}>
+    <Heart />
```
pdrbrnd: Why does Button have a Heart?

my-code/src/components/Button/styled.js
```diff
@@ -0,0 +1,71 @@
+// @flow
+import styled from "styled-components"
+
+type ThemeType = {
```
pdrbrnd: this can be moved to its own file, If is used more than once

my-code/src/components/Button/styled.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+const RegularButton = styled.button`
+  align-items: center;
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+
+  background: none;
+  background-color: ${(p: PropType) =>
```
pdrbrnd: yeah, I have to admit that flow here feels a bit strange, I think we're ok if we don't flow type the styled files

my-code/src/components/IconButton/index.js
```diff
@@ -0,0 +1,6 @@
+// @flow
```
pdrbrnd: I guess this wasn't used

my-code/src/constants/omdbApi.js
```diff
@@ -0,0 +1,6 @@
+const omdbApi = {
+  API_KEY: "a4e8e9b2",
```
pdrbrnd, line 2: careful with adding API keys to the repo.
take a look at [dotenv](https://www.npmjs.com/package/dotenv)

my-code/src/containers/App.js
```diff
@@ -0,0 +1,23 @@
...
+import { BrowserRouter, Route, Switch } from "react-router-dom"
+
+import Main from "./Main"
+import Detail from "./Detail"
+import TopBar from "../layout/topbar"
+
+import { Flex } from "../layout/Flex"
+
+const App = () => (
+  <BrowserRouter>
```
pdrbrnd: Maybe a 404?

my-code/src/containers/Detail.js
```diff
@@ -0,0 +1,214 @@
...
+type DirectorType = {
+  actor: string
+}
+
+type RatingsType = {
+  Source: string,
+  Value: string
+}
+
+class Detail extends React.Component<PropsType, StateType> {
```
pdrbrnd, line 50: What if the ID doesn't return a movie?

my-code/src/containers/Detail.js
```diff
@@ -0,0 +1,214 @@
...
+  constructor() {
+    super()
+    this.state = {
+      info: {},
+      favourite: false,
```
pdrbrnd: shouldn't we check if it's favorite before setting it to false?

my-code/src/containers/Detail.js
```diff
@@ -0,0 +1,214 @@
...
+      active: false
```
pdrbrnd: what's the active?
Can't we just use favorite?

That being sad, do we need the favorite key in the state or could we just use localstorage as the source-of-truth for favorites?

my-code/src/containers/Detail.js
```diff
@@ -0,0 +1,214 @@
...
+                  </div>
+                </div>
+              ))}
+              <Button
+                text="Add to favourites"
+                onClick={this.saveToFavourites}
+                active={active}
+              />
+            </Wrapper>
+          ) : null}
```
pdrbrnd: If it's a conditional render no need to use ternaries and return null. you could just `info.Ratings &&`

my-code/src/components/Button/index.js
```diff
@@ -0,0 +1,18 @@
...
+import React from "react"
+
+import RegularButton from "./styled"
+import Heart from "../Icons/Heart"
+
+type PropsType = {
+  text: string
+}
+
+const Button = ({ text, ...props }: PropsType) => (
```
danilowoz: Did you use `onChange`? If so, please create a type for it

my-code/src/components/Results.js
```diff
@@ -0,0 +1,51 @@
...
+}
+
+type MovieType = {
+  imdbID: string,
+  Title: string,
+  Year: string,
+  Poster: string
+}
+
+const Results = (props: PropsType) => {
```
danilowoz: Quick tip:
```jsx
const Results = ({ data }: PropsType) => {
```

Then you can remove the next line

my-code/src/components/Thumbnail/styled.js
```diff
@@ -0,0 +1,45 @@
...
+  theme: {
+    color: {
+      grey: {
+        grey: string
+      }
+    }
+  }
+}
+
+const SmallEmptyThumbnail = styled.div`
```
danilowoz: You can export directly the const here, instead of at the end of the file:

`export const SmallEmptyThumbnail = styled.div` `

my-code/src/containers/Detail.js
```diff
@@ -0,0 +1,214 @@
...
+      }),
+      () => localStorage.setItem("favourite", !favourite)
+    )
+  }
+
+  render() {
+    const { info } = this.state
+    const { active } = this.state
+
+    console.log(info)
```
danilowoz: Avoid leaving the consoles, is there any warning in the eslint?

danilowoz (changes requested):

danilowoz:
Thanks @isabelsoasa 🎉

PR #7, opened 2018-10-27
Challenge henrique

hjdesigner:
Hello, i'm Henrique. Nice to meet you! 
I am a Brazilian developer passionate about writing code for the web.
I always try to be in tune with what is happening in the front-end world and always willing to contribute to Open Source projects.

To know a little more than I do follow some links:
- https://henriquemelanda.com.br/
- https://codepen.io/hjdesigner/
- https://meu-curriculo.wedeploy.io/?hjdesigner
- https://github.com/hjdesigner/


my-code/.env
```diff
@@ -0,0 +1 @@
+API_IMDB=&apikey=296eb63f
```
danilowoz, line 1: You should never commit any env var, take a look at `dotenv` project

my-code/src/app.js
```diff
@@ -0,0 +1,170 @@
...
+        .then(response => response.json())
+        .then((data) => {
+          if (data.Response === 'False') {
+            this.setState({
+              statusEmpty: 'show'
+            })
+            return
+          }
+          if (isFavourites.length > 0) {
+            this.setState({
```
danilowoz, line 78: Remember that `this.setState` is asynchronous, is better to use only one `setState` by a method, it can getting some unexpected behavior. So you could do something like this:
```jsx
         this.setState({
            disableFavourites: !!isFavourites.length,
            movie: data,
            statusMovie: 'show',
            statusEmpty: 'hide'
          })
```

my-code/src/app.js
```diff
@@ -0,0 +1,170 @@
...
+      })
+    }
+    this.handleFavourites = (id) => () => {
+      const newFavourite = this.state.favourites
+      newFavourite.push({'imdb': id})
+      this.setState({
+        disableFavourites: true,
+        favourites: newFavourite
+      })
+      localStorage.setItem('favourites', JSON.stringify(this.state.favourites))
```
danilowoz, line 106: You could handle this in `componentDidUpdate`, so every new favorite update you can set it in `localStorate`

my-code/src/app.js
```diff
@@ -0,0 +1,170 @@
...
+    }
+  }
+  componentDidMount () {
+    const url = window.location.href
```
danilowoz, line 110: What about reuse the `this.handleGetMovie` method`

my-code/src/components/empty/empty.story.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import Empty from './index'
+
+const stories = storiesOf('Component Empty', module)
+
+stories.add('Empty Default', () => (
+  <div style={{ background: '#0a1014', fontFamily: 'Roboto', padding: 20, width: '100%', display: 'flex' }}>
```
danilowoz, line 8: You could create a component named Container or Wrapper, than it would avoid to duplicate so many code:
```jsx
const Wrapper = ({children}) => {
   return (<div style={ ...styles}>{children}</div>)
}

danilowoz:
Thanks for your submission

PR #10, opened 2019-07-11
Fernando aguilar

frclba:


movie-search/src/components/Movie.js
```diff
@@ -0,0 +1,128 @@
+import React from "react"
```
danilowoz, line 1: It seems great, but be careful about somethings: 
- The file is too large, try to split the component into small pieces;
- If so, it will avoid having too many `if/else`, only in this file there are about 9 check statements;

movie-search/src/components/Movie.js
```diff
@@ -0,0 +1,128 @@
...
+              />
+            ) : (
+              <Button
+                text="Add to favorites"
+                onClick={prop.toggleFavorite}
+                active={prop.favorite}
+              />
+            )}
+          </Wrapper>
+        ) : null}
```
danilowoz: It's not a big deal, but you don't need to return `null`.
If you check like `{prop.info.Ratings && /* */}` is enough

movie-search/src/components/Movie.js
```diff
@@ -0,0 +1,128 @@
...
+            </div>
+          </FlexLeft>
+
+          <FlexLeft pr={56}>
+            <Text100 grey mb={2}>
+              Director
+            </Text100>
+            <div>
+              {prop.info.Director
+                ? prop.info.Director.split(",").map(director => (
```
danilowoz: You could create a generic function for these three maps, like: 
```jsx
const renderInfos = item => (
    <Text100 key={item}>{item}</Text100>
)


// then

{prop.info.Genre ? prop.info.Genre.split(",").map(renderInfos) : "Unknow genre"}
```

movie-search/src/components/common/Label/index.js
```diff
@@ -0,0 +1,32 @@
+import React from "react"
+import PropTypes from "prop-types"
+
+import { Text100 } from "../Typography/styled"
+import { IMDB, Rotten, Label } from "./styled"
+import { Wrapper } from "../../Layout"
+
+export const IMDBLabel = ({ value }) => (
```
danilowoz, line 8: Prefer to create variables with a name more meaningful, avoid using services/modules for the name and think more what the component itself represents.

movie-search/src/components/common/Thumbnail/styled.js
```diff
@@ -0,0 +1,52 @@
...
+  border-radius: 2px;
+  background-color: ${({ theme }) => theme.color.grey.grey};
+`
+
+export const ThumbnailHolder = styled.a`
+  position: relative;
+  margin-bottom: 20px;
+`
+
+export const ThumbnailText = styled.div`
```
danilowoz: If it is a component that will handle a text/string, so use heading or paragraph, instead of `div`

movie-search/src/components/common/Typography/styled.js
```diff
@@ -0,0 +1,64 @@
+import styled, { css } from "styled-components"
+import { space } from "styled-system"
+
+const base = css`
```
danilowoz, line 4: Great 👍

movie-search/src/components/common/Typography/styled.js
```diff
@@ -0,0 +1,64 @@
...
+    text-decoration: none;
+  }
+
+  &:active {
+    color: ${({ theme }) => theme.color.grey.midgrey};
+    text-decoration: none;
+  }
+`
+
+const Text100 = styled.p`
```
danilowoz: In my opinion, these names are not meaningful enough, it would be better with you've used the font property as the name: like `<Smaller />`, `<SMall />`, so on...

movie-search/src/containers/MoviePage.js
```diff
@@ -0,0 +1,118 @@
...
+
+  getIsFavorite = () => {
+    const {
+      match: {
+        params: { id }
+      },
+      favorites
+    } = this.props
+
+    return favorites.filter(fav => fav === id).length > 0
```
danilowoz, line 37: Could be: `favorites.findIndex(fav => fav === id) !== -1`
Same result, differents approaches

movie-search/src/index.js
```diff
@@ -0,0 +1,21 @@
+import React from "react"
+import ReactDOM from "react-dom"
+
+import { createStore, applyMiddleware } from "redux"
+import { Provider } from "react-redux"
+import { createLogger } from "redux-logger"
```
danilowoz: It's recommended to use only for development environments, so it would great to create a different store for each environment.

movie-search/src/containers/MainPage.js
```diff
@@ -0,0 +1,86 @@
...
+          <form
+            style={{ width: "100%" }}
+            onSubmit={event => this.formSubmitted(event)}
+          >
+            <Input
+              width={1}
+              type="text"
+              value={searchTerm}
+              onChange={event =>
+                this.props.onSearchTermChanged(event.target.value)
```
danilowoz, line 34: Once you're only fetching after the submit, it could be a local state and send the value in the submit. That avoids unnecessary complexity.

movie-search/src/reducers/mainPage.js
```diff
@@ -0,0 +1,32 @@
+const initialState = {
+  searchTerm: "",
+  results: [],
+  loading: false
+}
+
+export default function mainPage(state = initialState, action) {
+  switch (action.type) {
+    case "SEARCH_TERM_CHANGED": {
```
danilowoz, line 9: A better approach is to save the actions name in constants.
https://redux.js.org/basics/actions#actions

movie-search/src/store.js
```diff
@@ -0,0 +1,51 @@
+import { searchMovies, searchSingleMovie, getFavorites } from "./API"
+
+const SEARCH_TERM_CHANGED = "SEARCH_TERM_CHANGED"
```
danilowoz: It should be declared in another place, in the reducer file or in a new document called actions.

movie-search/.env
```diff
@@ -0,0 +1 @@
+REACT_APP_API_KEY=a08065ae
```
danilowoz, line 2: You need to add these files in the `.gitignore`, it is not recommended to commit that

movie-search/README.md
```diff
@@ -0,0 +1,67 @@
+[![Build Status](https://travis-ci.org/fernand0aguilar/frontend-challenge.svg?branch=fernando-aguilar)](https://travis-ci.org/fernand0aguilar/frontend-challenge) 
+[![Maintainability](https://api.codeclimate.com/v1/badges/5e908ef8184fcc8707eb/maintainability)](https://codeclimate.com/github/fernand0aguilar/frontend-challenge/maintainability)
+
+***
+## About this project:
```
danilowoz, line 5: Great 👍

movie-search/README.md
```diff
@@ -0,0 +1,67 @@
...
+## Technical Details:
+
+**All the below technologies were used:**
+
+|Name|Description|Website| 
+|--|---|---|
+|CodeClimate| continuously measure and monitor code quality |[link](https://codeclimate.com/)|
+|Docker|  create, deploy, and run applications by using containers |[link](https://www.docker.com/)|
+|Eslint| The pluggable linting utility |[link](https://github.com/Significa/eslint-config-significa)|
+|Flow| A static type checker |[link](https://flow.org/)|
```
danilowoz: Sorry, but I didn't find any flow check

movie-search/README.md
```diff
@@ -0,0 +1,67 @@
...
+|Node| A JavaScript runtime server|[link](https://nodejs.org/en/)|
+|React| library for building user interfaces |[link](https://reactjs.org/)||
+|Selenium| framework for testing web applications. |[link](https://www.seleniumhq.org/)|
+|Styled System| Style props for rapid UI development |[link](https://styled-system.com/)|
+|Surge| A static web publishing |[link](https://surge.sh/)|
```
danilowoz, line 51: Nice, but where can I check the URL to your site?

movie-search/package.json
```diff
@@ -0,0 +1,35 @@
...
+		"styled-components": "^4.3.2",
+		"styled-system": "^3.1.11"
+	},
+	"scripts": {
+		"start": "react-scripts start",
+		"build": "react-scripts build",
+		"test": "react-scripts test --passWithNoTests",
+		"eject": "react-scripts eject"
+	},
+	"eslintConfig": {
```
danilowoz, line 25: You've used two approaches to set up the eslint: here in the `package.json` and using `.eslintrc`

danilowoz (commented):
Thanks for your submission @fernand0aguilar 🎉 
In general, your codebase looks great, but I left some thought about where you might improve your code. Some of those commentaries are just opinions and other ones are more about development conventions, readability, and maintainable issues.

However, the layout doesn't look really nice. There is some component without styles or wrong styling, and as you know, Significa is also a design agency. That's why we always care about delivering a website with amazing design and smooth user experience.

frclba:
Hello @danilowoz 😊
First of all, this is a great feedback! Thank you for taking a closer look.

I acknowledge the project mistakes and I appreciate your comments on the files. 
It's very handy and transparent to use github as a method of code review right? I love it.

I am improving on these suggestions and will come back by pushing new commits in no time.

If there's something yet to be said before I get back to fix them is Thanks to all Significa for the opportunity and attention, I've been growing a lot with this project.

PR #12, opened 2020-01-07
Joao Marques

marqu3s:
Hi!
I'm submitting my test.
It take so long because I'm actually studying React. So I had to do a lot of research.
I would really appreciate some feedback.


.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
```
danilowoz: Duplicated `editorconfig`, you have one on the root and another one on `my-code`

my-code/src/assets/icon-arrow-grey.svg
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
danilowoz: It would be great if all SVG are minified.

my-code/src/components/Header/index.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
```
danilowoz, line 1: Just a code style opinion: add a new line break between absolute and relatives imports. 
It will help you to see which modules are local or come from an external package.

my-code/src/components/Header/index.js
```diff
@@ -0,0 +1,11 @@
...
+import { Container } from './styles';
+import logo from '../../assets/logo.svg';
+
+export default function Header() {
+  return (
+    <Container>
+      <img src={logo} alt="Logo What's in" />
```
danilowoz, line 9: Great, you used the alt attribute on the image.

my-code/src/components/LikeButton/index.js
```diff
@@ -0,0 +1,67 @@
...
+import iconHeartWhite from './assets/icon-heart-white.svg';
+import iconHeartGrey from './assets/icon-heart-grey.svg';
+import iconHeartFull from './assets/icon-heart-full.svg';
+
+export default function LikeButton({ imdbID, showText, alwaysVisible }) {
+  const [liked, setLiked] = useState(false);
+  const dispatch = useDispatch();
+
+  // Redux state: get the favourites list.
+  const likes = useSelector(state => state.likesReducer);
```
danilowoz: It is a great idea to put all selectors on the reducer file, then if someday the "model" of reducer has changed, then it will be easier to update all selectors. Got it?

my-code/src/components/LikeButton/index.js
```diff
@@ -0,0 +1,67 @@
...
+
+  // Based on the favourites list, set the liked state.
+  useEffect(() => {
+    setLiked(likes.indexOf(imdbID) >= 0);
+  }, []);
```
danilowoz: Actually you have two dependencies on this effect: `likes` and `imdbID`.
Check your `eslint` rules and if your editor code is validating this kind of rule.

my-code/src/components/LikeButton/index.js
```diff
@@ -0,0 +1,67 @@
...
+      type: 'SET_LIKE',
+      imdbID,
+    });
+  }
+
+  return (
+    <Container>
+      <Button
+        type="button"
+        className={`heart-button ${liked || alwaysVisible ? 'visible' : ''} ${liked ? 'liked' : ''} ${showText && liked ? 'liked-red' : ''}`}
```
danilowoz: You could pass props to the component instead of creating a lot of className rules, it seems a little illegible. 
See more [here](https://www.styled-components.com/docs/basics#passed-props)

my-code/src/components/LikeButton/index.js
```diff
@@ -0,0 +1,67 @@
...
+        onClick={e => handleClick(e)}>
```
danilowoz: `onClick={handleClick}` also works, just a tip :)

my-code/src/components/LikeButton/index.js
```diff
@@ -0,0 +1,67 @@
...
+        <img
+          src={
+            (liked && iconHeartFull) ||
+            (showText ? iconHeartGrey : iconHeartWhite)
+          }
+          alt="Like it"
+        />
+        {showText && !liked && 'Add to favourites'}
```
danilowoz: I think it could be improved, you are checking the same value unnecessarily. 
Maybe: 
```js
showText && (liked ? "Added" : "Add to favourites")
```

my-code/src/components/MovieRating/index.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Container } from './styles';
+import logoImdb from '../../assets/logo-imdb.svg';
+import logoRotten from '../../assets/logo-rotten-tomatoes.svg';
+
+export default function MovieRating({ rating }) {
+  let logo;
```
danilowoz: There is nothing wrong here, but it could be a little bit more elegant, some ideas:
```js
  let logo = logoRotten;
  let logoCssClass = 'logo-rotten';

  if (rating.Source === 'Internet Movie Database') {
    logo = logoImdb;
    logoCssClass = 'logo-imdb';
  }
```

Or:

```js
  const isImdb = rating.Source === 'Internet Movie Database'
  
  const logo = isImdb ? logoImdb : logoRotten;
  const logoCssClass = isImdb ? 'logo-imdb' : 'logo-rotten';
```

Alright? What do you think?

my-code/src/components/Search/index.js
```diff
@@ -0,0 +1,117 @@
...
+  const [searchStr, setSearchStr] = useState('');
+
+  // Redux state: get the movies list.
+  const movies = useSelector(state => {
+    return state.movieReducer;
+  });
+
+  useEffect(() => {
+    setSearchStr(movies.searchStr);
+  }, []);
```
danilowoz: Again, you have a dependency on this effect.

my-code/src/components/Search/index.js
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * Fetches data from the API.
+   * @param {integer} currentPage
+   */
+  async function getData(currentPage) {
```
danilowoz: I would just update the function name to something more meaningful: `getMoveList` or `getSearchResult`. Reading this function I know what it does, but I don't know what kind of date it is.

my-code/src/components/Search/index.js
```diff
@@ -0,0 +1,117 @@
...
+    setIsLoading(true);
+    const searchResult = await fetchData(searchStr, currentPage, perPage);
```
danilowoz: Same thing here, I know what it does, but I don't which data it is.

my-code/src/components/Search/index.js
danilowoz: Also, make sure that currentPage has a default value.

my-code/src/components/Search/styles.js
```diff
@@ -0,0 +1,55 @@
...
+  border: 0;
+  color: ${variables.COLOR_DARK};
+  padding: 12px 12px 12px 40px;
+  width: 100%;
+
+  ::placeholder {
+    color: ${variables.COLOR_SECONDARY};
+  }
+
+  ${props =>
```
danilowoz, line 44: Nice, here's you have used props, check some component above to use this same approach.

my-code/src/components/Search/styles.js
```diff
@@ -0,0 +1,55 @@
...
+    props.isDisabled &&
+    css`
+      background-image: url(${magnifierDisabled});
+      background-color: ${variables.COLOR_DISABLED};
+      color: ${variables.COLOR_MIDGREY};
+
+      ::placeholder {
```
danilowoz, line 51: Shouldn't it be `&::placeholder`?

my-code/src/pages/Home/index.js
```diff
@@ -0,0 +1,124 @@
...
+                  </span>
+                </div>
+              </div>
+            </div>
+          ))}
+        </MoviesGrid>
+      )}
+
+      {!isLoading && movies.pageCount > 0 && (
+        <ReactPaginate
```
danilowoz, line 110: It would be nice if the pagination updates the browser history.

my-code/src/pages/Home/styles.js
```diff
@@ -0,0 +1,109 @@
+import styled from 'styled-components';
+import variables from '../../styles/variables';
+
+export const Container = styled.div`
+  ul.pagination {
```
danilowoz, line 5: Is it really necessary for the `ul` selector? The code seems "dirty" with this kind of selector.

my-code/src/pages/Movie/index.js
```diff
@@ -0,0 +1,132 @@
...
+      } else {
+        setNotFound(false);
+        setMovie(response.data);
+      }
+
+      setIsLoading(false);
+    };
+
+    fetchData();
+  }, []);
```
danilowoz: You have some missed dependencies here

my-code/src/store/modules/likes/reducer.js
```diff
@@ -0,0 +1,16 @@
+export default function LikesReducer(state = [], action) {
+  switch (action.type) {
+    case 'SET_LIKE':
+      const index = state.indexOf(action.imdbID);
```
danilowoz, line 4: You could simplify this part using `map`, `filter` and/or `reduce`, it will turn your code easier to read.

my-code/src/store/modules/movie/reducer.js
```diff
@@ -0,0 +1,18 @@
+export default function movieReducer(
+  state = { totalResults: 0, page: 0, pageCount: 0, searchStr: '', list: [] },
```
danilowoz, line 2: You might create a constant to this default value and use it again on [here](https://github.com/Significa/frontend-challenge/pull/12/files#diff-0895afad1f2e6842d750e76cb2655ecdR74) and [here](https://github.com/Significa/frontend-challenge/pull/12/files#diff-efa9ccd18515e4f1a53fb7bd651d61e3R23)

danilowoz (commented):
Your codebase seems nice, it has a folder structure pretty organized and it follows some good practices of development. 

Some point worth to mention: you've used async/await correctly and persist the data (but I wouldn't use a package for that, I would prefer to have more control of what and when persist the data).

--- 

Just walk through the comments and let me know if you have any doubts.

marqu3s:
Hi Danilo!
First, a big thank you for taking a time to write such valuable feedback!
I'll read carefully each one of your comments.
What happens now? The challenge ends here?
Is that ok if I make the adjustments, discuss somethings and send another PR when I'm done?

.editorconfig
marqu3s: I'll delete the one on the root.

my-code/src/assets/icon-arrow-grey.svg
marqu3s: Got a reduction from 16,3 KB to 5.08 KB (68.8% smaller) on the main assets folder and a reduction of 79% on the assets from the LikeButton component. Nice tip!

my-code/src/components/LikeButton/index.js
marqu3s: Got it and did it.

my-code/src/components/LikeButton/index.js
marqu3s: I had used an empty array to execute the effect only once as recommended by the useEffect documentation. The value of imdbID will not change over time and the component just have to know if the movie is already liked when it renders the first time to show an empty or a filled heart icon.

After adding the dependencies the effect is executed 10 times (1 time per movie on screen) after each like button is clicked on the search results page while using an empty array this doesn't happens.

Am I thinking the wrong way?

I've added the dependencies anyway since I adjusted my ESLINT and now it is auto-fixing this on save and it didn't broke the app.

my-code/src/components/LikeButton/index.js
marqu3s: That worked! Thanks!

my-code/src/components/LikeButton/index.js
marqu3s: Done! Much better now.

my-code/src/components/MovieRating/index.js
marqu3s: I'll stick with the second option you provided. It uses `const` and fewer lines.

my-code/src/components/Search/styles.js
marqu3s: It works with and without `&`.

my-code/src/pages/Home/index.js
marqu3s: What are the advantages of doing this since the last page visited is stored on the redux store/localStorage?

my-code/src/pages/Home/styles.js
marqu3s: Was used because I thought it would provide more meaning. If someone adds another `ul` in the container in the future, the browser will apply the pagination formatting to the new `ul` also (even if a class is provided). Agreed or do you still think it isn't necessary?

danilowoz:
Hey João, really nice!
Send an email to us addressing this PR on the body. Also, you can see more details about our career process [here](https://significa.co/careers/front-end).

I hope we both have learned something new from this review, and as we are done here I'm going to close this PR, alright?

PR #13, opened 2020-04-08
Lucas Nascimento

lucasnr:


src/pages/HomePage/MovieCard/styles.js
```diff
@@ -0,0 +1,57 @@
+import styled from 'styled-components';
+import { Link } from 'react-router-dom';
+
+export const Description = styled.div`
+	background-color: rgba(53, 63, 76, 0.9);
+	display: flex;
+	cursor: pointer;
+	flex-direction: column;
+	height: 100%;
+	margin-top: calc((100vw - 2rem) * 1.333);
```
danilowoz, line 10: Good use of relatives values as the unit of sizes and margins.

src/pages/HomePage/MovieCard/styles.js
```diff
@@ -0,0 +1,57 @@
+import styled from 'styled-components';
+import { Link } from 'react-router-dom';
+
+export const Description = styled.div`
+	background-color: rgba(53, 63, 76, 0.9);
```
danilowoz, line 5: Just a suggestion, it would be better if you create some kind of logic to order your style, like by alphabetic or group of rules. On my components, I prefer to create a structure like this:

```css
.foo {
  display: flex;
  flex-direction: column;
  
  /* empty line to split the group */
  font-size: 16px;
  color: inherit;

  ...other rules
}
```

Someday when you need to update some value, it will not take too much time looking for the rule and updating it. Also, there is an argument that ordering by alphabetic would avoid conflicts when there is more than one developer working on the project. So I would suggest picking the approach that fits better to your project.

src/pages/HomePage/MovieCard/styles.js
```diff
@@ -0,0 +1,57 @@
...
+	display: flex;
+	cursor: pointer;
+	flex-direction: column;
+	height: 100%;
+	margin-top: calc((100vw - 2rem) * 1.333);
+	padding: 0.5rem;
+	transition: 0.2s ease;
```
danilowoz, line 12: Prefer to define which rule you would like to do the transition. As you're making the transition of every property, you can face unexpected behaviours, but if you're sure of it, be more explicit and put `all` instead of nothing. 

```css
transition: all 200ms ease;
```

src/pages/HomePage/MovieCard/styles.js
```diff
@@ -0,0 +1,57 @@
...
+
+	@media (min-width: 430px) {
+		margin-top: 270px;
```
danilowoz, line 15: Pick only one unit to your values, I think that `rem` is a great choice.

src/pages/HomePage/MovieCard/styles.js
```diff
@@ -0,0 +1,57 @@
...
+	background-color: rgba(53, 63, 76, 0.9);
+	display: flex;
+	cursor: pointer;
+	flex-direction: column;
+	height: 100%;
+	margin-top: calc((100vw - 2rem) * 1.333);
+	padding: 0.5rem;
+	transition: 0.2s ease;
+
+	@media (min-width: 430px) {
```
danilowoz, line 14: Try to put all this kind of variables in a single place, it would be a `constant` file or even the theme of [Styled-components](https://styled-components.com/docs/api#themeprovider). Then you can reuse these values along of applications and make the breakpoints more solid.

src/components/Logo/index.js
```diff
@@ -0,0 +1,15 @@
...
+import logo from '../../assets/img/logo.svg';
+
+const Logo = styled.img`
+	margin-bottom: 1rem;
+	margin-right: auto;
+`;
+
+Logo.defaultProps = {
+	src: logo,
+	alt: `What's in's logo`,
```
danilowoz, line 12: 👍

src/pages/HomePage/index.js
```diff
@@ -0,0 +1,64 @@
...
+	const [results, setResults] = useState([]);
+	const [value, setValue] = useState('');
+
+	useEffect(() => {
+		if (value === '') {
+			setResults([]);
+			return;
+		}
+
+		searchMovies(value).then(({ data }) =>
```
danilowoz, line 28: Make sure to only request to the server when the user is ready. I mean, it's requesting on every `keydown` event, so there is a lot of unnecessary requests here. For instance, I've searched for `007` and the network has 3 requests, but only the last one is helpful. Did you get my point?

Screenshot 2020-04-08 at 11 09 01

To solve it, there are some common approaches to do it, like [debounce and throttle](http://loopinfinito.com.br/2013/09/24/throttle-e-debounce-patterns-em-javascript/). Try to think which one fits better for this problem.

src/pages/HomePage/index.js
```diff
@@ -0,0 +1,64 @@
...
+			setResults(data.results.filter((item) => item.poster_path))
+		);
+	}, [value]);
```
danilowoz, line 31: **Feature suggestion**: What about put the query search as a query string on URL? 

My navigation flow was like this:
1. I searched for `007` for example;
2. Clicked on the first movie, then I entered on the page movie;
3. Then I went back to the previous page (search page), and I noticed that my search wasn't there anymore; 
4. The user experience would be better to persist the last query search.

src/pages/HomePage/styles.js
```diff
@@ -0,0 +1,88 @@
+import styled from 'styled-components';
+
+import illustration from '../../assets/img/illustration-empty-state.png';
+
+export const Container = styled.main`
```
danilowoz, line 5: Great use of semantic tags, good one.

src/pages/HomePage/styles.js
```diff
@@ -0,0 +1,88 @@
...
+	align-items: center;
+	display: flex;
+	flex-direction: column;
+	justify-content: center;
+	margin: auto;
+	max-width: 1200px;
```
danilowoz, line 11: Mixed units, here is a value using `px`, above is `rem`.

src/pages/HomePage/styles.js
```diff
@@ -0,0 +1,88 @@
...
+	max-width: 440px;
+	width: 100%;
+
+	@media (min-width: 576px) {
+		background-size: cover;
+	}
+`;
+
+export const Title = styled.h2`
+	font-size: 1.5rem;
```
danilowoz, line 43: Take a look at `ThemeProvider` of styled-components, I think it will help you to make the style more solid.

src/pages/MoviePage/AddToFavouriteButton/index.js
```diff
@@ -0,0 +1,19 @@
+import React, { useState, useCallback } from 'react';
+import { IoIosHeartEmpty, IoIosHeart } from 'react-icons/io';
+
+import { Container } from './styles';
+
+export default function AddToFavouriteButton() {
+	const [favourite, setFavourite] = useState(false);
+
+	const handleAddToFavourites = useCallback(() => {
```
danilowoz, line 9: Great use of `useCallback`, because you have sure that `favourite` has changed before update it again. Another approach could be:

```ts
const handleAddToFavourites = () => {
   setFavourite(prev => !prev);
}
```

src/pages/MoviePage/index.js
```diff
@@ -0,0 +1,121 @@
...
+								{genres.map((item) => (
+									<li key={`genre-${item.id}`}>{item.name}</li>
+								))}
+							</ul>
+						</List>
+						<List>
+							<Subtitle>Director</Subtitle>
+							<ul>
+								{credits.crew
+									.filter((item) => item.job === 'Director')
```
danilowoz, line 108: Take a look at method `find`, I think it would be helpful here

src/services/api.js
```diff
@@ -0,0 +1,13 @@
+import axios from 'axios';
+
+const API_KEY = '16da3c7c6882c8c6fe2e82b63b8f8823';
```
danilowoz, line 3: Never push API keys or secret keys to Github, take a look at `dotenv`.
It's not a big deal here, as it's a challenge using a public API, but in a real project, it would be a serious issue.

src/App.js
```diff
@@ -0,0 +1,14 @@
...
+import { BrowserRouter as Router } from 'react-router-dom';
+
+import Routes from './routes';
+import GlobalStyle from './styles/GlobalStyle';
+
+export default function App() {
+	return (
+		<Router>
+			<Routes />
+			<GlobalStyle />
```
danilowoz, line 11: Place the `GlobalStyle` for first, because right now styles are rendering in a wrong order, see:

CeEV3NfWqxpvN1ADohxPGwNe

Because styled-component will render the components and only when all this job is done, it will add the global style. But the global one must be the first processed style, then you can overwrite the global style on each component if it is needed of course.

danilowoz (commented):
Thank you for submitting our frontend challenge.

src/pages/HomePage/MovieCard/styles.js
lucasnr: I always try to do it in alphabetic order. I don't know if i forgot it somewhere but thanks for pointing it out.

src/services/api.js
lucasnr: Thanks for pointing that out. I had that in mind when i was coding, and only pushed it because it was not that a big deal as you said.

PR #14, opened 2020-05-20
Lourenco

costa86:


my-code/src/components/FetchMovie.js
```diff
@@ -0,0 +1,59 @@
+import React, { useState, useEffect } from "react";
+import { useParams } from 'react-router';
+
+export function FetchMovie() {
+    const [data, setData] = useState({});
+    const { id } = useParams();
+    const [loading, setLoading] = useState(true);
+
+    async function getData() {
```
danilowoz: There are a couple of things that worth to mention:
- Async functions should be performed inside a `useEffect` function, and not in the root of components, it can create some unexpected behavior in your component, [here a great example](https://www.robinwieruch.de/react-hooks-fetch-data) of how to use `useEffect` with API fetches.
- Split the concerns: there are too many things happening here, try to split the main functions in small ones;
- Unnecessaries `let`: most of the variable values haven't changed over the function, so every variable could be a `const` which avoid unwanted mutability;
- Constants values: I would recommend getting all static values (like `apiKey` and `url`) from a specific place (maybe `src/constants`), then you can consume these values throughout applications from a single place.
- Sensitive data: information like `apiKey` should be hidden from public access, so I recommend use https://github.com/motdotla/dotenv for this case.

my-code/src/components/Movies.js
```diff
@@ -0,0 +1,50 @@
...
+
+    const [query, setQuery] = useState("");
+    const [disabled, setDisabled] = useState(true);
+    const [search, setSearch] = useState(false);
+
+    let apiKey = "798e7a67";
+
+    let min = 3;
+
+    function queryMovie() {
```
danilowoz: Native methods are strongly not recommended to update or get DOM elements in React applications. As React uses `state` to trigger renders, this information it just got from DOM can be stale (from a previous render). If it is really necessary to interact with dom elements, take a look at this [section](https://reactjs.org/docs/refs-and-the-dom.html) in React documentation.

But for this case, React `state` will provide what you need to achieve the same result, even to get the value from input, [see more here](https://reactjs.org/docs/forms.html).

my-code/src/components/Movies.js
```diff
@@ -0,0 +1,50 @@
...
+        setQuery(q.value);
+        setSearch(true);
+        results.innerHTML = `Results for ${q.value}:`;
+        document.title = `${q.value} | MOVIES SEARCH`;
+
+    }
+
+    useEffect(() => {
+        document.getElementById("q").focus();
+    })
```
danilowoz: It's important to set the dependencies on `useEffect`.
https://dev.to/spukas/4-ways-to-useeffect-pf6

danilowoz (commented):
Thank you for submitting to our front-end challenge.

costa86:
Hi, @danilowoz . Thanks a lot for the suggestions! I did some changes. I look forward to getting your review.

my-code/src/components/AllMoviesSearch.js
```diff
@@ -0,0 +1,72 @@
...
+    // Custom hook to make API requests (disabled while query === null)
+    const [data, loading] = useFetch(url, query);
+    //
+    const [disabled, setDisabled] = useState(true); // Submit button's status
+    let res;
+    const q = document.getElementById("q");
+
+    // Keeps focus on input field
+    useEffect(() => {
+        document.getElementById("q").focus();
```
danilowoz: You'll be able to achieve the same behavior using `useRef` from React API, instead of using native methods which aren't the safest path in a React application.

my-code/src/components/AllMoviesSearch.js
```diff
@@ -0,0 +1,72 @@
...
+    // Enables & makes the API request
+    const search = () => {
+        setQuery(q.value);
+        document.title = `${q.value.toUpperCase()} | ${APP_NAME}`;
+    }
+    //
+
+    // Enables/disables submit button & shows chars counter
+    function enableBtn() {
+        let q = document.getElementById("q");
```
danilowoz: I think the main issue here is how you're getting the value of the input, try to follow this pattern:
```jsx
const App = () => {
  const [inputValue, setInputValue] = useState('');

  return (
    <>
      <input
        value={inputValue}
        onChange={(event) => setInputValue(event.target.value)}
      />
      <button disabled={inputValue.length > 0}>Search</button>
    </>
  );
};

```
Following it, you'll be able to remove all JS native methods and you'll make sure that there is a [single source of true](https://fjorgedigital.com/insights/blog/the-philosophy-of-react-a-single-source-of-truth/).

my-code/src/components/AllMoviesSearch.js
```diff
@@ -0,0 +1,72 @@
...
+        const hint = document.getElementById("hint");
+        (q.value.length >= min) ? setDisabled(false) : setDisabled(true);
+        hint.innerHTML = `${q.value.length}/${min}`;
+    }
+    //
+
+    // Actual content to be rendered, based on API request status
+    switch (loading) {
```
danilowoz, line 48: It isn't a good idea works with a mutation in a [functional component](https://programmingwithmosh.com/react/react-functional-components/) in React, the recommended pattern for this case is using [early returns](https://www.robinwieruch.de/conditional-rendering-react), so it would be something like this:

```jsx
if(loading) {
  return <p>Loading...</p>
}

if(noData) {
  return <p>Search again...</p>  
}

// From here I'm sure that it isn't loading and there are data enough to render the list
return <List data={data} />
```

danilowoz (commented):
Hey, nice to see that you've done some progress here. 
I just left a couple of comments again, hope it's clear and helps you.

costa86:
Hi again, @danilowoz ! Just did some changes based on your feedback. Thanks again!

danilowoz:
Hey @costa86, it seems you've done a great progress here!
From now on, I'd recommend you follow the next steps [here](https://significa.co/careers/front-end). Feel free to ask anything which is not clear yet.

Thanks

costa86:
Hi, @danilowoz .I have just submitted the application form!

PR #16, opened 2020-06-07
Fernando gomes

FernandoMGomes:
Hey, this is m challenge submission.

### Run Project

`npm install`

`npm run build`

`npm run dev`


my-code/src/components/addMovieButton.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import heartGrey from '../assets/icons/icon-heart-grey.svg'
+import heartFull from '../assets/icons/icon-heart-full.svg'
+
+const searchInput = (props) => {
```
danilowoz, line 5: There are some things that worthy of mentioning here:
- All React components should be in PascalCase, which let clear that is a class or a component;
- It's interesting to use some kind of prop type check, read more about: PropTypes, Typescript or Flow;
- I can see there is some variation of the same component, but it doesn't need to repeat the HTML three times, as you can work with props and conditional render, let me try to give you an example:
```jsx
const Text = ({ children, withIcon }) => {
  return <p className={withIcon ? "has-icon" : "normal"}>
       {withIcon && <i className="icon" />}
       {children}
    </p>
}
```

my-code/src/components/searchInput.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+const searchInput = (props) => {
+    return (
+        <div>
+            <input type="text"
+                placeholder="Search movies..."
+                onChange={(e) => { props.handleChange(e) }}
```
danilowoz, line 8: Just a tip, the following code does the same thing:
```jsx
onChange={(e) => { props.handleChange(e) }}
onChange={props.handleChange}
```

my-code/src/containers/Home.js
```diff
@@ -0,0 +1,105 @@
...
+    const value = event.target.value;
+    this.setState({ searchValue: value,loading: true, error: '' },
+      () => {
+        this.fetchMovies(value);
+      }
+    );
+  }
+
+  fetchMovies(value) {
+    fetch(`http://www.omdbapi.com/?apikey=a6487992&s=${value}` , {
```
danilowoz, line 39: Never push API keys or secret keys to Github, take a look at dotenv.
It's not a big deal here, as it's a challenge using a public API, but in a real project, it would be a serious issue.

my-code/src/containers/Home.js
```diff
@@ -0,0 +1,105 @@
...
+      data: [],
+      searchValue: '',
+      loading: false,
+      error: '',
+      addedMovies: [],
+    }
+  }
+
+  componentDidMount() {
+    if(localStorage.getItem('addedMovies')) {
```
danilowoz, line 22: Same static values like `addedMovies` in constant above, it's a great practice to know that this value will not change and to share between functions and scopes.

Read more: https://www.robinwieruch.de/javascript-naming-conventions

my-code/src/containers/Home.js
```diff
@@ -0,0 +1,105 @@
...
+      }
+    })
+    .catch((error) => {
+      console.log(error)
+    })
+  }
+
+  handleAddButton = (id) => {
+    console.log(this.state)
+    if(this.state.addedMovies.includes(id)) {
```
danilowoz, line 60: I would recommend splitting this function into small pieces, there are too many things happening here. For example, the local storage handle might be a util.

my-code/src/containers/Home.js
```diff
@@ -0,0 +1,105 @@
...
+      });
+    }
+  }
+
+  renderContent() {
+    if (this.state.loading) {
+      return (<EmptyStateManager state={'loading'} />)
+    }
+    if (this.state.data.length > 0) {
+      return (
```
danilowoz, line 79: Just a tip, once there is a return into a conditional, the `else` statement is not necessary.

my-code/src/containers/Home.js
```diff
@@ -0,0 +1,105 @@
...
+      const parseData = JSON.parse(retrievedData);
+      this.setState({ addedMovies: parseData})
+    }
+  }
+
+  handleChange(event) {
+    const value = event.target.value;
+    this.setState({ searchValue: value,loading: true, error: '' },
+      () => {
+        this.fetchMovies(value);
```
danilowoz, line 33: Make sure to only request to the server when the user is ready. I mean, it's requesting on every `keydown` event, so there is a lot of unnecessary requests here. For instance, I've searched for `007` and the network has 3 requests, but only the last one is helpful. Did you get my point?

To solve it, there are some common approaches to do it, like [debounce and throttle](http://loopinfinito.com.br/2013/09/24/throttle-e-debounce-patterns-em-javascript/). Try to think which one fits better for this problem.

my-code/src/styles/_colors.scss
```diff
@@ -0,0 +1,9 @@
+$yellow: #FF9F1C;
```
danilowoz, line 1: Naming colors variables is always a nightmare, so let me suggest a new pattern we've been working internally:
```scss
// Grayscale
$g100: #eee;
$g200: #222;
$g250: #444;
$g300: #999;

// Variation of brand color
$b100: #353f4c;
$b200: #1b2329;

$error: red;
$sucess: green;
```

Using this approach the variable name becomes more predictable, and the name doesn't carry about its value, for example, what about if the `$red` is not red anymore?

my-code/package.json
```diff
@@ -0,0 +1,70 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/ReactJSResources/react-webpack-babel"
+  },
+  "author": "ReactJSResources",
+  "license": "MIT",
+  "homepage": "https://github.com/ReactJSResources/react-webpack-babel#readme",
+  "dependencies": {
+    "@babel/cli": "^7.8.3",
```
danilowoz, line 26: There are a lot of dependencies here, that are dev dependencies actually. Of course, the final bundle will not have these packages, but it's important for the next developer to understand what is dev stack used in the project.

danilowoz (commented):
Thank you for submitting our frontend challenge.

I just left some comments on your code, it's some suggestions and tips, and not mandatory changes. Feel free to ask whatever question you have about my review. From now on, I'd recommend you follow the next steps [here](https://significa.co/careers/front-end).

FernandoMGomes:
Hey Danilo,
first let me say is not normal to get a code review of a challenge, I really appreciate you taking the time, also, is good to see a bit of your code culture.
I will respond and resolve each one of the issues, this is more of a general response.
This challenges always get me confuse, I never know how deep should I go or the code culture of the company, so there are always some misses.

my-code/package.json
FernandoMGomes: True, I just use a boilerplate for the project, if this was for a production project the webpack and the dependencies had to be done by hand removing all the crap.

my-code/src/styles/_colors.scss
FernandoMGomes: Looking at the invision page, colors had names, just thought it was something to follow.

my-code/src/containers/Home.js
FernandoMGomes: I though I wrapped this on setTimeout :/

my-code/src/containers/Home.js
FernandoMGomes: For legibility my current employer prefers this way, after some time you kinda get used to it.

my-code/src/containers/Home.js
FernandoMGomes: I feel ashamed of this rookie mistake...

danilowoz:
Thanks @FernandoMGomes for your words.
I'm going to close your PR here, but feel free to ask any questions you may have.

PR #17, opened 2020-06-16
Ana monteiro

anamonteiro430:


movies_challenge/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "movies_challenge",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@testing-library/jest-dom": "^4.2.4",
+    "@testing-library/react": "^9.3.2",
+    "@testing-library/user-event": "^7.1.2",
```
danilowoz, line 8: Try to don't mix up development dependencies between regular dependencies. Actually, it's not a big deal as the final bundle will not have these packages, but it's a good practice to keep the dependencies split between `devDependencies` and `dependencies`.

movies_challenge/src/App.css
```diff
@@ -0,0 +1,38 @@
+.App {
+  text-align: center;
+}
+
+.App-logo {
+  height: 40vmin;
+  pointer-events: none;
+}
+
+@media (prefers-reduced-motion: no-preference) {
```
danilowoz, line 10: Good one!

movies_challenge/src/App.css
```diff
@@ -0,0 +1,38 @@
+.App {
+  text-align: center;
+}
+
+.App-logo {
+  height: 40vmin;
```
danilowoz, line 6: I just learned a new here, I haven't used `vmin` or `vmax` before.

movies_challenge/src/App.css
```diff
@@ -0,0 +1,38 @@
...
+}
+
+@media (prefers-reduced-motion: no-preference) {
+  .App-logo {
+    animation: App-logo-spin infinite 20s linear;
+  }
+}
+
+.App-header {
+  background-color: #282c34;
```
danilowoz, line 17: I noticed you're using sass with variables in the project, I think you left one behind.

movies_challenge/src/App.css
```diff
@@ -0,0 +1,38 @@
...
+  min-height: 100vh;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  font-size: calc(10px + 2vmin);
+  color: white;
```
danilowoz, line 24: It's a good idea to turn every color in a variable, even the most common values like white and black because it might change throughout the development process.

movies_challenge/src/Components/Home.js
```diff
@@ -0,0 +1,78 @@
...
+import Movies from "./Movies";
+import { Loading } from "./Loading";
+
+function Home(props) {
+  const api_key = REACT_APP_API_KEY;
+  const [searchParam, setSearchParam] = useState("");
+  const [moviesList, setMoviesList] = useState();
+  const [error, setError] = useState(false);
+  const [loading, setLoading] = useState(false);
+  const [empty, setEmpty] = useState(true);
```
danilowoz, line 14: As long as too many states handling the same goal here, I'd suggest to turn it in a [`useReducer`](https://reactjs.org/docs/hooks-reference.html#usereducer) or use React Redux. It will deliver some important performance improvements and the legibility for your code.

movies_challenge/src/Components/Home.js
```diff
@@ -0,0 +1,78 @@
...
+
+  const searchMovie = (e) => {
+    setError(false);
+    e.preventDefault();
+    if ((searchParam === "") | (searchParam === undefined)) {
+      setEmpty(true);
+    } else {
+      setError(false);
+      setLoading(false);
```
danilowoz, line 23: Here's an interesting example of why the `useReducer` or Redux is important. 
It might trigger unnecessary renders, as two states have been updating in a row, so your interface might encounter some expected behaviors, like showing a loading with an error message.

With the approaches that I've mentioned above, you'll be able to manage the both states at once, then you'll make sure what is the current state of your application.

movies_challenge/src/Components/Home.js
```diff
@@ -0,0 +1,78 @@
...
+        .then((res) => {
+          if (res.data.Response === "True") {
+            setMoviesList(res.data.Search);
+            setEmpty(false);
+            setLoading(false);
+          } else {
+            setError(true);
+          }
+        })
+        .catch((err) => console.log("NOT VALID STRING"));
```
danilowoz, line 36: As you already dealing with errors, it would be a good idea to set the error state here too.

movies_challenge/src/Components/Home.js
```diff
@@ -0,0 +1,78 @@
...
+      <form onSubmit={searchMovie}>
+        <input
+          name="search"
+          type="text"
+          placeholder="Search"
+          value={searchParam}
+          onChange={handleChange}
+        />
+      </form>
+      {loading && !error ? (
```
danilowoz, line 55: Instead of using nested ternary which doesn't have good legibility, I'd recommend to use early return to render this section, like:

```jsx
const renderResults = () => {
  if(loading) {
    return <p> Loading</p>
  }

  if(error) {
    return <p>Error</p>
  }

  return <div>Result</div>
}

return <div>
{renderResults()}
</div>
 ```

Read more about the different conditional render [here](https://www.robinwieruch.de/conditional-rendering-react)

File: movies_challenge/src/Components/MoviePage.js [hunk 10]
```diff
@@ -0,0 +1,160 @@
...
+import heart_white from "./../assets/icon-heart-full.svg";
+import logo_imdb from "./../assets/logo-imdb.svg";
+import { Loading } from "./Loading";
+
+function MoviePage(props) {
+  let id = props.match.params.id;
+  const api_key = REACT_APP_API_KEY;
+  const [loading, setLoading] = useState(false);
+  const [movie, setMovie] = useState({});
+  const [hover, setHover] = useState({
```
danilowoz, line 17:
```js
const [hover, setHover] = useState(false)
```

It might be enough. Actually, you don't need to put every time an object into the state.

movies_challenge/src/Components/MoviePage.js
```diff
@@ -0,0 +1,160 @@
...
+        <Loading />
+      ) : (
+        <>
+          <Link
+            to={"/"}
+            className="arrow-back"
+            onMouseOver={handleMouseOver}
+            onMouseOut={handleMouseOut}
+          >
+            {hover.arrow ? (
```
danilowoz, line 78: You could use conditional in attributes as well, then it avoids repeats code, like:
```jsx
<img
  src={hover.arrow ? arrow_white : arrow_grey}
  width="30"
  alt="arrow-back"
  onMouseEnter={handleMouseOver}
  onMouseLeave={handleMouseOut}
/>
```

movies_challenge/src/Components/MoviePage.js
```diff
@@ -0,0 +1,160 @@
...
+                    <img src={logo_imdb} />
+                  </div>
+                  <div className="movie-info--imdb_rating">
+                    <p>{movie.imdbRating} / 10</p>
+                  </div>
+                </div>
+
+                {fav.toggled ? (
+                  <a className="fav-added" onClick={removeFav}>
+                    <img src={heart_white} width="30" alt="arrow-back" />
```
danilowoz, line 121: Alt attributes should have a value that describes the image content, like "Back icon" instead of an id format.

movies_challenge/src/Components/Movies.js
```diff
@@ -0,0 +1,46 @@
...
+        }}
+      >
+        <div id="movie">
+          {props.movie.Poster === "N/A" ? (
+            <img src={placeholder} alt="poster" width="180" />
+          ) : (
+            <img src={props.movie.Poster} alt="poster" width="180" />
+          )}
+
+          {hover ? (
```
danilowoz, line 29: I'm pretty sure that is possible to deal only with CSS, which avoids adding a ton of event listeners at the DOM.

movies_challenge/src/styles/styles.scss
```diff
@@ -0,0 +1,311 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@100;300;400;500;700;900&display=swap");
+
+$red: #ff4040;
```
danilowoz, line 3: Naming colors variables is always a nightmare, so let me suggest a new pattern we've been working internally here at Significa:
```scss
// Grayscale
$g100: #eee;
$g200: #222;
$g250: #444;
$g300: #999;

// Variation of brand color
$b100: #353f4c;
$b200: #1b2329;

$error: red;
$sucess: green;
```

Using this approach the variable name becomes more predictable, and the name doesn't carry about its value, for example, what about if the `$red` is not red anymore?

movies_challenge/.gitignore
```diff
@@ -0,0 +1,26 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
+
+# dependencies
+/node_modules
+/src/settings
```
danilowoz, line 5: It a good practice to never push API keys or secret keys to Github, take a look at dotenv project which is the right way to manage those keys.

danilowoz (commented):
Thank you for submitting in our frontend challenge, there are a lot of interesting things here. 🚀 

I just left some comments on your code, it's some suggestions and tips, and not mandatory changes. Feel free to ask whatever question you have about my review. From now on, I'd recommend you follow the next steps [here](https://significa.co/careers/front-end).

anamonteiro430:
Thank you so much Danilo , useful stuff! Thanks for taking the time

danilowoz:
Thanks again @anamonteiro430, I'm going to close your PR here, but feel free to ask any questions you may have.

PR #19, opened 2020-12-04
Challenge pedro

Pedro-Mac:
Hi!

I am opening a pull request for challenge review :smile:

I hope I _broke a leg_

Cheers!


client/.eslintcache
```diff
@@ -0,0 +1 @@
...\Users\\Admin\\projects\\frontend-challenge\\client\\src\\components\\HeartFavorites\\index.jsx",[],"C:\\Users\\Admin\\projects\\frontend-challenge\\client\\src\\components\\Arrow\\index.jsx",[],"C:\\Users\\Admin\\projects\\frontend-challenge\\client\\src\\components\\Loader\\index.jsx",[],"C:\\Users\\Admin\\projects\\frontend-challenge\\client\\src\\components\\MovieNotFound\\index.jsx",[],"C:\\Users\\Admin\\projects\\frontend-challenge\\client\\src\\views\\SingleMovie\\ArrowBack\\index.jsx",[],{"ruleId":"73","replacedBy":"74"},{"ruleId":"75","replacedBy":"76"},{"ruleId":"77","severity":1,"message":"78","line":44,"column":6,"nodeType":"79","endLine":44,"endColumn":18,"suggestions":"80"},{"ruleId":"73","replacedBy":"81"},{"ruleId":"75","replacedBy":"82"},"no-native-reassign",["83"],"no-negated-in-lhs",["84"],"react-hooks/exhaustive-deps","React Hook useEffect has a missing dependency: 'searchError'. Either include it or remove the dependency array.","ArrayExpression",["85"],["83"],["84"],"no-global-assign","no-unsafe-negation",{"desc":"86","fix":"87"},"Update the dependencies array to be: [searchError, searchText]",{"range":"88","text":"89"},[1654,1666],"[searchError, searchText]"]
```
danilowoz: Usually, this file shouldn't be in the repository, as it only matters for development purpose on your machine. So consider adding this file into gitignore, in this repo.

client/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "client",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@testing-library/jest-dom": "^5.11.4",
```
danilowoz: Neither `@testing-library/**` nor `dotenv` is used in the client, but only on build and development time. Probably these dependencies came with the create-react-app when you started your project, that's fine. But actually, these dependencies are only used on development mode, so it would be `devDependencies` instead of `dependencies`.

client/src/App.jsx
```diff
@@ -0,0 +1,44 @@
+import React, { useState, useEffect } from "react";
+import { BrowserRouter, Route, Switch } from "react-router-dom";
+import ls from "local-storage";
+import Home from "./views/Home";
+import SingleMovie from "./views/SingleMovie";
+import "./App.scss";
+
+import Logo from "./images/2.Logos/logo.svg";
```
danilowoz: In a React application, each variable that is in PascalCase we assume to be a React component or a Javascript class object, everything else must be in camel case, from props to other variables.

This's one of the most important convension on the React community, be careful while naming your variables.

client/src/App.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+function App() {
+  const [favorites, setFavorites] = useState([]);
+
+  useEffect(() => {
+    const favoriteMovies = ls.get("favoriteMovies") || [];
```
danilowoz: Another important convention is creating constants throughout your project. Here's a good example, as the key of `localstorage` is used more than one time and **its value should not change**, you could create a constant to assign it, like:

```js
const LOCALSTORAGE_KEY = "__ favoriteMovies__"
```

Also, be careful to not repeat the naming, like you have two "favoriteMovies" things, with different values. 

Read more here: https://www.javascripttutorial.net/es6/javascript-const

client/src/App.jsx
```diff
@@ -0,0 +1,44 @@
...
+    setFavorites(favoriteMovies);
+  }, []);
+
+  return (
+    <div className="App col-sm-11 col-md-9">
+      <img src={Logo} alt="logo" className="Logo"></img>
```
danilowoz: In order to improve the accessibility, consider provide a better alternative text for images, try to describe what is there or what is written. 

---

Image elements has no children, so self-closed it, like: 
```jsx
<img src={logoSrc} alt="What's in" />
```

client/src/components/Arrow/index.jsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import { Link } from "react-router-dom";
+import ArrowBackGrey from "../../images/1.Icons/icon-arrow-grey.svg";
+import ArrowBackWhite from "../../images/1.Icons/icon-arrow-white.svg";
+
+const Arrow = ({ className, type, alt, to, onMouseEnter, onMouseLeave }) => {
+  return (
+    <div className={className}>
+      <Link to={to} onMouseEnter={onMouseEnter} onMouseLeave={onMouseLeave}>
```
danilowoz, line 9: Mouseover effects might easily achieve without any javascript. Using CSS will deliver a smooth user experience with no pain for the browser.

client/src/components/HeartFavorites/index.jsx
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import HeartIconGrey from "../../images/1.Icons/icon-heart-grey.svg";
+import HeartIconWhite from "../../images/1.Icons/icon-heart-white.svg";
+import HeartIconFull from "../../images/1.Icons/icon-heart-full.svg";
+
+const HeartFavorites = ({ style, className, type, alt }) => {
+  return (
+    <img
+      src={
+        type === "white"
```
danilowoz: Nested ternary is an anti-pattern and should be avoided since it's difficult to maintain and read. One way to solve this specific case is creating a function with earlies return, where you're able to set how many conditionals you want, like: 
```js
const getImageSource = () => {
 if (type === "white") {
   return HeartIconWhite
 }

 if (type === "grey") {
   return HeartIconGrey
 }

 return HeartIconFull
}

return <img src={getImageSource()} />
```

client/src/components/HomeText/index.jsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+const HomeText = () => {
+  return (
+    <div className="d-flex flex-column align-items-center mt-2">
+      <h3 className="font-medium-high font-weight-normal text-white">
+        Don't know what to search?
+      </h3>
+      <p style={{ color: "#7a8c99" }}>Here's an offer you can't refuse</p>
```
danilowoz, line 9: Sass could take care of styles and color variables, avoid spreading inline-style in the project.

client/src/components/Loader/style.scss
```diff
@@ -0,0 +1,25 @@
+.lds-dual-ring {
+  display: inline-block;
+  width: 128px;
+  height: 128px;
+  margin-top: 12rem;
```
danilowoz, line 5: Choose only one CSS unit and use through the project. Relative units are always better than pixels, so have in mind to use one of them, never both.

client/src/components/MovieNotFound/style.scss
```diff
@@ -0,0 +1,14 @@
+@import "../../styles/master";
+.movie-not-found {
+  p {
+    color: #fff;
+    a {
+      color: $color-negative;
+      text-decoration: none;
+
+      &:hover {
```
danilowoz, line 9: Hover's style must be wrapper by a media-query, because many mobile devices emulate hovering when the user performs an inconvenient long tap

```css
@media (hover: hover) {
  a:hover {
    background: yellow;
  }
}
```

Read more https://developer.mozilla.org/en-US/docs/Web/CSS/@media/hover

client/src/components/SearchBar/index.jsx
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+// import MagnifierDisabled from "../../images/1.Icons/icon-magnifier-disabled.svg";
```
danilowoz: Be careful about forgetting commented code in your repository, it's not a good practice.

client/src/views/Home/MovieItems/styles.scss
```diff
@@ -0,0 +1,45 @@
...
+        color: $shade-default;
+      }
+
+      img {
+        width: 1.25rem;
+        height: auto;
+      }
+
+      div {
+        p {
```
danilowoz, line 32: The deeply nested class name is not a good practice in CSS, it's better to use a lot of specific class names than creating so many levels on selectors.

client/src/views/Home/index.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+const Home = ({ favorites }) => {
+  const [searchText, setSearchText] = useState("");
+  const [moviesList, setMoviesList] = useState([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [searchError, setSearchError] = useState(false);
+
+  useEffect(() => {
+    if (searchText) {
+      const httpRequest = setTimeout(async () => {
```
danilowoz, line 52: The purpose of this `setTimeout` is to create a denounce, isn't it? If so, it's a good way to avoid excessive requests.

client/src/views/Home/index.jsx
```diff
@@ -0,0 +1,67 @@
...
+        setIsLoading(true);
+        const { data } = await axios.get(
+          `http://www.omdbapi.com/?s=${searchText}&apikey=${process.env.REACT_APP_OMDB_API_KEY}`,
+        );
+        console.log("http request");
```
danilowoz: Use console.log only for development purpose, do not deliver an application throwing consoles in production.

client/src/views/Home/index.jsx
```diff
@@ -0,0 +1,67 @@
...
+import "./styles.scss";
+import axios from "axios";
+
+const Home = ({ favorites }) => {
+  const [searchText, setSearchText] = useState("");
+  const [moviesList, setMoviesList] = useState([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [searchError, setSearchError] = useState(false);
+
+  useEffect(() => {
```
danilowoz, line 50: This `useEffect` is doing too many things at the same time: 
- Requesting to the API;
- Get loading and error state;
- Debounce;
- Listening to a keydown event;

Consider breaking it down in the smallest pieces, which will do only one thing. It will improve the readability, testability and maintainability of the function.

danilowoz (commented):
Thank you for submitting in our frontend challenge, there are a lot of interesting things here. 🚀

Some general thoughts:
- We wanted to see your CSS skills, so for another challenge avoid UI libraries, like bootstrap;
- In general, your code is legible, but you need to have in mind what the purpose of which function and split into smallest pieces in order to reuse in other parts of the code;
- Consider deploying it in someplace (vercel, netlify, heroku), it would be a great plus.

Thanks again

client/src/views/Home/index.jsx
Pedro-Mac: Hi Danilo! 😄 

Would you recommend that I implement multiple `useEffect` functions, each one responsible for less effects, instead of a single one responsible for many effects?

danilowoz: Hey
Not really, I'd recommend creating some kind of abstraction, for example:
```js
const searchQueryDebounced = useDebounce(searchQuery)

useEffect(() => { 
 setLoading(true)
 const movies = getMovies(searchQueryDebounced)
 // etc
}, [searchQueryDebounced])
```

client/.eslintcache
Pedro-Mac: Hi Danilo!

I will try to go point by point make changes it in the project. I will reply on each point in order to track what I have done. That way I will be sure of what I have changed and if you feel curious and like giving feedback, I will highly appreciate - I am looking to improve every aspect of my code.

Added lintcache to dotenv

client/src/App.jsx
Pedro-Mac: Changed images variable names to camelCase.

client/src/App.jsx
Pedro-Mac: created constant  for local storage key

client/src/components/SearchBar/index.jsx
Pedro-Mac: Hi Danilo! I was assuming that here you are referring to non-descriptive comments which are not helpful for me and for other people to understand the code, am I right?

client/src/views/Home/index.jsx
Pedro-Mac: Hi Danilo! Yes, that was exactly my goal, to make the request only some time the user has finished typing

client/src/views/Home/index.jsx
Pedro-Mac: Removed logs

client/src/views/Home/index.jsx
Pedro-Mac: Abstractions created for the effects in the home and single movie views

client/src/components/HeartFavorites/index.jsx
Pedro-Mac: Refactored nested ternaries into functions with switch statements

client/src/components/SearchBar/index.jsx
danilowoz: Sorry for the late answer, I was on vacation.

Actually, I was talking about this SVG's import which is commented line and you're not using it anywhere

PR #22, opened 2021-01-02
Isabel moreira

michkin:
Hello!
This is my take on the movie app challenge.

A few notes:
- A few commits are regarding my (failed 😢) attempts at deploying the app with github pages directly from this repository. I ended up deploying with Netlify and Gitlab (something I used in the past). You can check it at [whats-in.netlify.app](https://whats-in.netlify.app/). 
- I didn't hide my API key right away, so you'll have access to it in the first commits. I wanted to keep the app simple, but in a normal situation, I would do it right away. 
- Also, a few commits have to do with some issues with having more than one `.gitignore` in the repository. 
 
Thanks! 😊
Isabel


my-code/significa/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "homepage": "http://michkin.github.io/",
+  "name": "significa",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@emotion/react": "^11.1.2",
+    "@emotion/styled": "^11.0.0",
+    "@reach/router": "^1.3.4",
+    "@testing-library/jest-dom": "^5.11.6",
```
danilowoz, line 10: These packages would make more sense as devDependencies.

my-code/significa/src/components/Buttons.js
```diff
@@ -0,0 +1,127 @@
...
+      onMouseEnter={() => setHeart(heartWhite)}
+      onMouseLeave={() => setHeart(heartGrey)}
+    >
+      <HeartIcon
+        src={heart}
+        alt="heart icon"
+        className="btn__icon"
+        data-testid="favNo"
+      />
+      {isSmall ? (
```
danilowoz, line 70: Nested ternary condition is never a good idea, instead try to turn in a render function, like: 
```js
const renderText = () => {
  if (isSmall) {
    return "Like"
  }

  if (isMediumLarge) {
    return "Add to favourites"
  }

  return null
}
```

my-code/significa/src/components/Buttons.js
```diff
@@ -0,0 +1,127 @@
...
+
+import heartWhite from "../assets/icons/icon-heart-white.svg";
+import heartGrey from "../assets/icons/icon-heart-grey.svg";
+import heartFull from "../assets/icons/icon-heart-full.svg";
+import arrowWhite from "../assets/icons/icon-arrow-white.svg";
+import arrowGrey from "../assets/icons/icon-arrow-grey.svg";
+
+/* General button styling */
+const Button = styled.button`
+  position: relative;
```
danilowoz, line 13: Just a tip: try to organize better the properties think in "blocks" of properties. For example, put all similar properties together and then divide them with an empty line, like

```css
display: flex;
align-items: center;
justify-content: center;

padding: 0;
margin: 5px 16px 0 0;
height: inherit;
border-radius: 4px;

font-size: 16px;
font-weight: bold;
cursor: pointer;
```

my-code/significa/src/components/Buttons.js
```diff
@@ -0,0 +1,127 @@
...
+
+const ButtonFavNo = () => {
+  const [heart, setHeart] = useState(heartGrey);
+  const isSmall = useMediaQuery({ minWidth: 500, maxWidth: 599 });
+  const isMediumLarge = useMediaQuery({ minWidth: 600 });
+
+  return (
+    <BtnFavNo
+      className="btn"
+      onMouseEnter={() => setHeart(heartWhite)}
```
danilowoz, line 61: This kind of effect could be done using CSS, controlling the visibility of two images. UI events on javascript should be used only necessary, CSS transitions are way better in terms of performance.

my-code/significa/src/components/Labels.js
```diff
@@ -0,0 +1,77 @@
...
+`;
+const Logo = styled.img`
+  padding: 8px;
+  border-bottom-left-radius: 4px;
+  border-top-left-radius: 4px;
+`;
+const Rating = styled.div`
+  font-size: 1rem;
+  padding: 8px;
+  border: 1px #353f4c solid;
```
danilowoz, line 31: `emotion` has support for custom themes, consider using it in order to keep all style guide values in a single place.
https://emotion.sh/docs/theming

my-code/significa/src/components/Labels.js
```diff
@@ -0,0 +1,77 @@
...
+// Label with logo and rating
+const Label = styled.div`
+  position: relative;
+  max-width: fit-content;
+  height: 100%;
+  max-height: 30px;
+  display: flex;
+  flex-flow: row nowrap;
+  align-content: stretch;
+  margin: 5px 1rem 0 0;
```
danilowoz, line 16: Avoid mixing units, my suggestion is to pick only one approach and use throughout the project, `rem` would be the best choice BTW.

my-code/significa/src/components/Labels.js
```diff
@@ -0,0 +1,77 @@
...
+  border-left: none;
+  line-height: 0.8;
+
+  @media only screen and (min-width: 768px) {
+    line-height: 1.5;
+  }
+`;
+
+const LabelLogo = ({ imdb, rotten, rating }) => {
+  const logo = imdb ? logoImdb : rotten ? logoRotten : null;
```
danilowoz, line 43: Two points which worths to mention here:
- Avoid nested conditionals;
- This props approach wouldn't scale in a real application, what would happen if it needs to support more 3 other logos? 

Take a look at following approach:
```jsx
// Component
const Logo = ({ variant, ...props }) => {
  const options = {
    imdb: logoImdb,
    rotten: logoRotten,
    netflix: logoNetflix,
    hbo: logoHbo
  }

  const currentLogo = options[variant]

  return current ? <img src={currentLogo} : null
}

// Using it
<Logo variant="imdb" />
```

Does it make sense? Can you see how it make the component more flexible?

my-code/significa/src/components/layout.css
```diff
@@ -0,0 +1,63 @@
+/* for every device */
```
danilowoz, line 1: It isn't common to use two approaches of styling at the same project for a countless reason, but the main ones would be the bundle optimization and developer experience, I mean, if I wanted to style a component, which approach should I use? 

So in general, CSS-in-js is a great choice in a React application, it's preferable to use `emotion` so.

my-code/significa/src/pages/Home.js
```diff
@@ -0,0 +1,190 @@
...
+      setEmpty(true);
+    } else {
+      setEmpty(false);
+    }
+    return setQuery(e.target.value.trim());
+  };
+
+  /* fetch list of movies from the api when query changes */
+  useEffect(() => {
+    let mounted = true;
```
danilowoz, line 34: I didn't get the reason for this, which kind of problem did you encountered here?

my-code/significa/src/pages/Home.js
```diff
@@ -0,0 +1,190 @@
...
+    };
+
+    const requestUrl = encodeURI(
+      `https://www.omdbapi.com/?s=${query}&apikey=${API_KEY}`
+    );
+    setIsLoading(true);
+    getList(requestUrl);
+
+    return () => (mounted = false);
+  }, [query]);
```
danilowoz, line 58: Make sure to only request to the server when the user is ready. I mean, it's requesting on every keydown event, so there is a lot of unnecessary requests here. For instance, I've searched for 007 and the network has 3 requests, but only the last one is helpful. Did you get my point?

To solve it, there are some common approaches to do it, like `debounce` and `throttle`. Try to think which one fits better for this problem.

my-code/significa/src/pages/Home.test.js
```diff
@@ -0,0 +1,122 @@
...
+  localStorage.setItem("whatsInFavourites", remFav);
+
+  // search movie "city lights" again
+  fireEvent.change(input, { target: { value: "city lights" } });
+
+  await waitFor(() => {
+    // check if movie has a full heart on the card, it should not, as it is a favourite
+    expect(screen.queryByTestId(movieId)).not.toBeInTheDocument();
+  });
+});
```
danilowoz, line 122: Awesome! 🚀

my-code/significa/src/pages/Movie.js
```diff
@@ -0,0 +1,211 @@
...
+
+  /* set variables that use state in the api call */
+  const [error, setError] = useState(null);
+  const [isLoading, setIsLoading] = useState(false);
+  const [movie, setMovie] = useState([]);
+
+  /* making favourites persistent (using localStorage) */
+  let favourites = localStorage.getItem("whatsInFavourites");
+
+  const [favourite, setFavourite] = useState(() => {
```
danilowoz, line 24: This isn't necessarily wrong, but the initial state of the state should not take care of this kind of logic but receives a value, not more than that. My suggestion would be:
- Create a util file to manage the localstorage, in order to avoid repeating code;
- There you'll be able to create abstraction around the localstorage and make it easier to work;
- So `useState` would receive a single value from this util:
```js
const { isItemFavorite } = useFavorite() // abstraction over localstorage
const [favourite, setFavourite] = useState(isItemFavorite(movieId))
```

Does it make sense?

my-code/significa/src/pages/Movie.js
```diff
@@ -0,0 +1,211 @@
...
+      return favourites.includes(params.movieId);
+    } else {
+      localStorage.setItem("whatsInFavourites", []);
+      return false;
+    }
+  });
+
+  /* functions that deal with favourites persistence */
+  const addFavourites = () => {
+    let newFav = localStorage.getItem("whatsInFavourites");
```
danilowoz, line 36: This could be replaced for the util function mentioned above.

danilowoz (commented):
Thank you for submitting our frontend challenge, there are a lot of interesting things here! 😄 

michkin:
Thank you so much @danilowoz, for reviewing my code! Especially since you're not hiring, it's very kind of you to take the time to make these corrections and suggestions.
I will study them better, and (if you're available, of course) I might pick your brain a little bit... 😊

danilowoz:
You're welcome @michkin! I'm going to close this PR, but feel free to reach me out if you have any doubt.

PR #29, opened 2022-07-10
Vinicios silva

VinewZ:
Hi Significa!

Found the challenge at: https://github.com/felipefialho/frontend-challenges and finished it.

Thx for making it available.


my-code/src/Components/MovieCard/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { useMovies } from '../../Contexts/MoviesData';
+import { FavouriteIcon, MovieCardContainer, MovieDetails, MovieInfo, MoviePoster } from './styles';
+import fullHeartIcon from '/icons/icon-heart-full.svg';
+import whiteHeartIcon from '/icons/icon-heart-white.svg';
+
+interface MovieCardProps {
+    Poster: string
+    Title: string
+    Year: string
+    imdbID: string
```
TGlide: Prefer lower-case props, upper-case is normally reserved for components, e.g. `MovieCard`

my-code/src/Components/MovieCard/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export function MovieCard({ imdbID, Poster, Title, Year }: MovieCardProps) {
+
+    const navigate = useNavigate()
+    const { isFavourite, addToFavourites } = useMovies()
+    const { getMovieInfo } = useMovieDetails()
+
+    function handleClick() {
+        navigate(`/movie/${imdbID}`)
+        getMovieInfo(imdbID)
```
TGlide: The page `MovieDetails` already runs `getMovieInfo`, so you shouldn't need to run this. You only need to change your useEffect in `MovieDetails` to: 
```
useEffect(() => {
    getMovieInfo(id!);
  }, [id]);
```

my-code/src/Components/MovieCard/styles.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export const MovieCardContainer = styled.div`
+    width: fit-content;
+    height: fit-content;
+    position: relative;
+`;
+
+export const MoviePoster = styled.img`
+    position: relative;
+    width: 180px;
```
TGlide: Since the Container size is dependent on the MoviePoster size (since the others are absolutely positioned elements), wouldn't it make sense to use rem here? Otherwise, if the user changes the default font-size, this may happen:
CleanShot 2022-07-13 at 16 21 21

TGlide: A good article about pixels vs rem: https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/

my-code/src/Components/MoviesList/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+    const { movies, isLoading } = useMovies()
+
+    return (
+        <MoviesListContainer>
+            {isLoading ? (
+                <>
+                    <Skeleton width={180} height={240} baseColor={theme.colors.midGrey} highlightColor={theme.colors.dark} />
+                    <Skeleton width={180} height={240} baseColor={theme.colors.midGrey} highlightColor={theme.colors.dark} />
+                    <Skeleton width={180} height={240} baseColor={theme.colors.midGrey} highlightColor={theme.colors.dark} />
+                </>
```
TGlide: Only 3 skeletons can look a bit off, specially on big viewports, because of the spacing, meaning the position of the skeletons won't reflect the position of the movies. What you could do, is use a display flex instead of grid (specially since your cards have a fixed width) with flex-wrap, and put more skeletons.

Going from this:
CleanShot 2022-07-13 at 16 31 31
To this:
CleanShot 2022-07-13 at 16 30 37

Quick tip for reusing multiple components too:

```
const range = (start: number, end: number) => {
  return Array.from({ length: end - start }, (_, i) => start + i);
};

// ...
{isLoading ? (
        range(0, 10).map((index) => {
          return (
            <Skeleton
              width={180}
              height={240}
              baseColor={theme.colors.midGrey}
              highlightColor={theme.colors.dark}
            />
          );
        })
// ...
```

my-code/src/Components/SearchMovieInput/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+import magnifierIcon from '/icons/icon-magnifier-grey.svg';
+
+export function SearchMovieInput() {
+
+    const { getMoviesData } = useMovies()
+    const [searchTerm, setSearchTerm] = useState('')
+
+    useEffect(() => {
+        getMoviesData(searchTerm)
+    }, [searchTerm])
```
TGlide, line 14: Maybe use a debounce to avoid triggering so many searches

my-code/src/Components/SearchMovieInput/styles.tsx
```diff
@@ -0,0 +1,24 @@
+import styled from 'styled-components';
+import { theme } from '../../GlobalStyles/Theme';
+
+export const SearchMoviesInputContainer = styled.div`
+    width: 100%;
+    height: 44px;
```
TGlide: Since this contains text, you should also consider if using px here is worthwhile (I exaggerated and put in 120px as the root font size):
CleanShot 2022-07-13 at 16 33 56

my-code/src/Contexts/MovieDetails.tsx
```diff
@@ -0,0 +1,55 @@
...
+    Title: string
+    Ratings: [{ Source: string, Value: string }]
+    Plot: string
+    Actors: string;
+    Genre: string;
+    Director: string
+    Poster: string
+}
+
+const MovieDetailsContext = createContext({} as MovieDetailsContextProps)
```
TGlide: This can cause errors, e.g. someone trying to call getMovieInfo, and receiving an error because undefined is not callable.

```suggestion
const MovieDetailsContext = createContext<MovieDetailsContextProps>({
  getMovieInfo: async () => {
    return;
  },
  movieInfo: {},
  isLoading: false,
});
```

TGlide: `movieInfo` is empty in this case, resulting in a type error. You could make all `MovieInfoProps` values be potentially undefined, or you could say that `movieInfo` can be undefined/null. Your call!

my-code/src/Contexts/MovieDetails.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export function MovieDetailsProvider({ children }: MovieDetailsProviderProps) {
+
+    const [movieInfo, setMovieInfo] = useState({} as MovieInfoProps)
```
TGlide:
```suggestion
      const [movieInfo, setMovieInfo] = useState<MovieInfoProps>({});
```
Same thing here, casting can lead to errors

my-code/src/Contexts/MovieDetails.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const [isLoading, setIsLoading] = useState(false)
+
+    async function getMovieInfo(imdbID: string) {
+        setIsLoading(true)
+        const response = await axios.get(`https://www.omdbapi.com/?apikey=23fc3dfd&i=${imdbID}`)
```
TGlide, line 67: What if an error occurs?

my-code/src/Contexts/MoviesData.tsx
```diff
@@ -0,0 +1,95 @@
...
+    Year: string;
+    imdbRating: string;
+    imdbID: string;
+}
+
+interface FavouriteMoviesProps {
+    imdbID: string;
+}
+
+const MoviesDataContext = createContext({} as MoviesDataContextProps)
```
TGlide: Same here

my-code/src/Contexts/MoviesData.tsx
```diff
@@ -0,0 +1,95 @@
...
+export function MoviesDataProvider({ children }: MoviesDataProviderProps) {
+
+    const [movies, setMovies] = useState<MovieProps[]>([])
+    const [favouriteMovies, setFavouriteMovies] = useState<FavouriteMoviesProps[]>(
+        JSON.parse(localStorage.getItem('favouriteMovies') || '[]')
+    )
+    const [isLoading, setIsLoading] = useState(false)
+
+    async function getMoviesData(search: string) {
+        if (search.length > 0 && search.length < 3) return;
```
TGlide: There is a movie called Pi, that this search wouldn't find: https://www.imdb.com/title/tt0138704/?ref_=nv_sr_srsg_8

my-code/src/Contexts/MoviesData.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+        setIsLoading(true)
+        if (search.length === 0) return (
+            setMovies([]),
+            setIsLoading(false)
+        )
```
TGlide: You have a line already that checks if search.length is less than zero. Shouldn't you include this check there? Unless you always want to do setIsLoading(false) with invalid search lengths... In that case:
```suggestion
        if (search.length >= 0) {
            setMovies([]),
            setIsLoading(false)
            return
        }

        setIsLoading(true)
```

my-code/src/Contexts/MoviesData.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+        const response = await axios.get(`https://www.omdbapi.com/?apikey=23fc3dfd&s=${search}`)
```
TGlide: You shouldn't include hardcoded API keys. That way, anyone with access to your github repo can exploit that key for personal use. You could use environment variables to avoid this.

my-code/src/Contexts/MoviesData.tsx
```diff
@@ -0,0 +1,95 @@
...
+        if (response.data.Response === 'False') return (
+            setMovies([]),
+            setIsLoading(false)
+        )
```
TGlide: This return is unusual. Why not do:
```suggestion
        if (response.data.Response === 'False') {
            setMovies([]),
            setIsLoading(false)
            return
        }
```
Or even: 
```suggestion
         if (response.data.Response === 'False') {
            setMovies([]),
            return setIsLoading(false)
        }
```

my-code/src/Contexts/MoviesData.tsx
```diff
@@ -0,0 +1,95 @@
...
+        setMovies(response.data.Search)
+        setIsLoading(false)
+    }
+
+    function addToFavourites(movieID: string) {
+        if (favouriteMovies.some(movie => movie.imdbID === movieID)) {
+            return setFavouriteMovies(favouriteMovies.filter(movie => movie.imdbID !== movieID))
+        }
+        setFavouriteMovies([{ imdbID: movieID }, ...favouriteMovies])
```
TGlide: This is a bit hard to read. Maybe do this:
```suggestion
        const isFavourited = favouriteMovies.some(movie => movie.imdbID === movieID)
        
        if (isFavourited) {
            return setFavouriteMovies(favouriteMovies.filter(movie => movie.imdbID !== movieID))
        }
        setFavouriteMovies([{ imdbID: movieID }, ...favouriteMovies])
```

Oh wait! You already have a function that checks if a movie is favourited:
```suggestion
        if (isFavourite(movieID)) {
            return setFavouriteMovies(favouriteMovies.filter(movie => movie.imdbID !== movieID))
        }
        setFavouriteMovies([{ imdbID: movieID }, ...favouriteMovies])
```

my-code/src/Contexts/MoviesData.tsx
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+
+    function isFavourite(movieID: string) {
+        return (favouriteMovies.some(movie => movie.imdbID === movieID))
+    }
+
+    useEffect(() => {
+        localStorage.setItem('favouriteMovies', JSON.stringify(favouriteMovies))
+    }, [favouriteMovies])
```
TGlide: You could include this in your addToFavourites function, without needing to rely on an useEffect

my-code/src/Pages/MovieDetails/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+export function MovieDetails() {
+
+    const { isFavourite, addToFavourites } = useMovies()
+    const { movieInfo, getMovieInfo, isLoading } = useMovieDetails()
+    const { id } = useParams()
+    const navigate = useNavigate()
+
+    useEffect(() => {
+        !movieInfo.imdbID && getMovieInfo(id!)
+    }, [id])
```
TGlide:
```suggestion
     useEffect(() => {
        getMovieInfo(id!);
     }, [id]);
```

my-code/src/Pages/MovieDetails/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+                            </TechnicalInfoContainer>
+                        </MovieInfoContainer>
+                        <MoviePosterContainer>
+                            <MoviePoster src={movieInfo.Poster} alt={movieInfo.Poster} />
+                        </MoviePosterContainer>
+                    </>
+                )
+            }
+        </MovieDetailsContainer>
+    );
```
TGlide:
```suggestion
 if (isLoading) {
    return (
      <MovieDetailsContainer>
        <Skeleton
          width={"85vw"}
          height={"95vh"}
          baseColor={theme.colors.midGrey}
          highlightColor={theme.colors.dark}
        />
      </MovieDetailsContainer>
    );
  }

  if (!movieInfo.imdbID) return null;

  return (
    <MovieDetailsContainer>
      <MovieInfoContainer>
        <ArrowIcon
          src={arrowIcon}
          alt="click to go back"
          onClick={() => navigate("/")}
        />
        <InfoHeader>
          <Runtime>{movieInfo.Runtime}</Runtime>
          <Year>{movieInfo.Year}</Year>
          <Rated>{movieInfo.Rated}</Rated>
        </InfoHeader>
        <MovieTitle>{movieInfo.Title}</MovieTitle>
        <RatingsContainer>
          {movieInfo.Ratings.slice(0, 2).map((rating, index) => (
            <Rating key={index}>
              <RatingLogo
                src={
                  rating.Source === "Internet Movie Database"
                    ? imdbLogo
                    : rottenTomatoesLogo
                }
                alt={rating.Source}
              />
              <p>{rating.Value}</p>
            </Rating>
          ))}
          <AddToFavouritesButton
            isFavourite={isFavourite(movieInfo.imdbID)}
            onClick={() => addToFavourites(movieInfo.imdbID)}
          >
            <img src={heartIcon} alt="Add to favourites" />
            <p>
              {isFavourite(movieInfo.imdbID) ? "Added" : "Add to favourites"}
            </p>
          </AddToFavouritesButton>
        </RatingsContainer>
        <TechnicalInfoContainer>
          <TechnicalInfo>
            <TechnicalInfoTitle>Cast</TechnicalInfoTitle>
            {movieInfo.Actors.split(",").map((actor, index) => (
              <p key={index}>{actor}</p>
            ))}
          </TechnicalInfo>
          <TechnicalInfo>
            <TechnicalInfoTitle>Genre</TechnicalInfoTitle>
            {movieInfo.Genre.split(",").map((genre, index) => (
              <p key={index}>{genre}</p>
            ))}
          </TechnicalInfo>
          <TechnicalInfo>
            <TechnicalInfoTitle>Director</TechnicalInfoTitle>
            {movieInfo.Director.split(",").map((director, index) => (
              <p key={index}>{director}</p>
            ))}
          </TechnicalInfo>
        </TechnicalInfoContainer>
      </MovieInfoContainer>
      <MoviePosterContainer>
        <MoviePoster src={movieInfo.Poster} alt={movieInfo.Poster} />
      </MoviePosterContainer>
    </MovieDetailsContainer>
  );
```

TGlide (commented):
Thank you for your submission! 😄 

I've written down some comments and suggested changes. Feel free to discuss them with me!

my-code/src/Components/MovieCard/index.tsx
VinewZ: Is it possible if the api response is an object with capitalized properties? 
I know that in this case we dont have access to the api code, but wouldnt it be easier if this change was made in the api itself so the properties would be lowercase?

image

my-code/src/Contexts/MoviesData.tsx
VinewZ: Line 49 was to "limit" the requests, but i learned today about debounce (thank you) so, i agree that it is useless.
Your suggestion on line 50~ wouldnt work (typo?), since:
````
if (search.length >= 0) {
            setMovies([]),
````
Would always reset the movies array each time the user searches for something.

I guess a little change would be a better solution:
````
if (search.length === 0) {
            setMovies([]),
            setIsLoading(false)
            return
        }
````

File: my-code/src/Contexts/MoviesData.tsx (hunk 13 above)
Comment (VinewZ) [reply my-code/src/Contexts/MoviesData.tsx]: I totally agree, but since it doesnt have any personal data and this is an open source project, where i expect people to see and run it (e.g. for a code review), I intended to leave it like this.

If you have any solution for a situation like this, i would like to know.

File: my-code/src/Components/MovieCard/index.tsx (hunk 1 above)
Comment (TGlide) [reply my-code/src/Components/MovieCard/index.tsx]: It is possible, since the only place you use `MovieCard` is on `MoviesList`. The refactor would be simple:
```
// Before
<MovieCard
	  key={movie.imdbID}
	  Poster={movie.Poster}
	  Title={movie.Title}
	  Year={movie.Year}
	  imdbID={movie.imdbID}
/>

// After
<MovieCard
	  key={movie.imdbID}
	  poster={movie.Poster}
	  title={movie.Title}
	  year={movie.Year}
	  imdbID={movie.imdbID}
/>
 ```
 
Now, if you were using it like this before:
```
<MovieCard
	  key={movie.imdbID}
	  {...movie}
/>
```
Things would be a bit more complicated. And I think we can come to a better, clearer solution:

Going to `Contexts/MoviesData`, you have an interface called `MovieProps`. This naming leads me to think that there is a component called `Movie`, and that these are its props. Since its not the case, instead, it could be called `Movie`, or `MovieData` or `MovieDatum`. Following this logic, `MoviesDataContextProps` could be `MoviesDataContext`/`MoviesDataContextType`/`MoviesDataContextT`, and `FavouriteMoviesProps`... well maybe it is not needed even:

```
// Before
const [favouriteMovies, setFavouriteMovies] = useState<FavouriteMoviesProps[]>
// After
const [favouriteMovies, setFavouriteMovies] = useState<Pick<Movie, 'imdbID'>>
```

This has the benefit of not having to create a new type, and it actually reflects the type of Movie. Meaning that if we change imdbID inside Movie to be, say, a number, due to API changes, we don't need to update in two places.

With that in mind, we can finally come to `MovieCard` again, and refactor it's type:
```
type MovieCardProps = Pick<Movie, "imdbID" | "Poster" | "Year" | "Title">;
```
With that same benefit of depending on the original type. But the props are still upper-case... In that case, I offer two alternatives:
```
type MovieCardProps = {
  imdbID: Movie['imdbID']
  title: Movie['Title']
  year: Movie['Year']
  poster: Movie['Poster']
}
// or
type MovieCardProps = {
  data: Movie;
}
```
That way the actual props are lower-case, while maintaining the Movie type as is, meaning only Movie accurately reflects the API. The component doesn't need to know what the API format is.

my-code/src/Contexts/MoviesData.tsx
TGlide: Yes, that was a typo, sorry! Your solution is better 🙂

my-code/src/Contexts/MoviesData.tsx
TGlide: You already have the live demo (which is great, thanks for including that!), and you could include steps to get a token on the README. But I agree that for this use-case, it's totally fine. The worst that could happen is someone overusing the key, then the key being revoked.

But in truth, unless you used a proxy, or something like Next.JS API routes to proxy the request, anyone can see the request by inspecting the network requests on the live demo, so even using environment variables here is exploitable.

PR #30, opened 2023-03-07
[Chore] Update challenge requirements and info

ItzaMi:
Here's a new iteration for the frontend challenge.
The most important differences are:
- there are no design files (we don't want to overcomplicate the work)
- removed any mention to React (considering our experiments, maybe we're ok with any framework?)
- clarified requirements (since we no longer have a design to show)

Let's discuss it! 🚀 

EDIT: 
Would it make sense to have drawings like this just to explain the intended layout?

Screenshot 2023-03-07 at 17 30 49

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,21 +4,31 @@
 
 ## About the challenge
 
-This is a pretty simple test, in which you should be able to write a React app, using the provided screens as reference. You’ll need to make requests to an API to get some JSON content, filter the data and print it on view.
+This challenge encompasses the following skills:
 
-You can use any boilerplate/approach you prefer, but try to keep it simple. We encourage you to use your favourite tools and packages to build a solid React application.
+1. Writing a solid application
+2. Making API calls
+3. Filter data and shape it to display on the UI
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,21 +4,31 @@
...
   _Think about loading state and movie not found_
 
-The purpose of this challenge is to evaluate your HTML, CSS, Javascript and React API skills and overall approach while composing a simple web app given a set of screens and an API feed. We will also assess the HTML, CSS, and JS output.
+The purpose of this challenge is to evaluate your HTML, CSS, Javascript and API skills and overall approach while composing a simple web app given a set of screens and an API feed. We will also assess the HTML, CSS, and JS output.
+
+### What is expected
+
+- A functional app with the two screens (`Home` and `Details`);
+- A clean layout and design;
+- A way to tag movies as _favourites_ (and having them persist would be great!);
```
dfalmeida, line 31: Shouldn't be `having them persisted would be great`

ItzaMi: I don't think it really matters. Seems to be ok both ways

dfalmeida:
Also regarding the drawing, I don't think its needed as the text is pretty self explanatory

PR #32, opened 2023-05-22
Frontend Challenge Solution

fernanda-freitas:
This PR contains the entire solution for the Significa's frontend challenge.


my-code/whats-in/src/index.scss
```diff
@@ -0,0 +1,513 @@
+$dark: #0A1014;
```
pdrbrnd, line 1: while I honestly like the straightforward approach of having a single css file for the entire app (and kudos for the BEM methodology), this makes it a bit harder to review and to fully understand of what's what – given it's completely decoupled from the markup it affects and a somewhat big file.

another problem with an approach like this is the undesired consequence that might happen when changing something in the CSS file. For instance, one developer gets a ticket to slightly update the style of certain screen but, by doing so, it will unintentionally affect some unrelated parts of the app that are using the same class.

A more paradigmatic approach in react apps and similar modern frameworks would be to use components to achieve what you're doing with classes and keep the style very tightly coupled with the markup. As an example, instead of having a `.state` class you could have a `<State>` component that would contain the styling in itself and accept the image, title and description as props.

my-code/whats-in/src/images/illustration-notfound-state@2x.png
pdrbrnd, line 1: nice work! love the vertigo reference :)

my-code/whats-in/src/components/Notfound/index.jsx
```diff
@@ -0,0 +1,11 @@
+import notFoundImage from '../../images/illustration-notfound-state@2x.png'
+
+export default function NotFound() {
+    return (
+        <div className='state'>
+            <img className='state__image--not-found' src={notFoundImage} alt="Empty search" />
```
pdrbrnd, line 6: the `alt` tag should be an image replacement on screen readers, so using a value like "Empty search" might be a bit confusing (what is an image of an empty search?).
I guess two valid approaches here would be:
- an empty alt tag: this image being purely decorative we could assume that it doesn't add much to the screen and simply tell the screen reader to ignore it.
- a description of the image: we are telling visually impaired people what we are seeing (e.g.: "A silhouette of a man falling down in a red spiral")

my-code/whats-in/src/components/Empty/index.jsx
```diff
@@ -0,0 +1,18 @@
...
+
+export default function Empty( ) {
+const navigation = useNavigate();
+
+    const handleNavigation = () => {
+        navigation('/')
+    }
+
+    return (
+        <div onClick={handleNavigation} className='state'>
```
pdrbrnd, line 12: the accessibility (a11y) of this approach could be improved. 
Ideally, you should strive to use the appropriate tags. In this case, given it's simply navigation, an anchor tag (`<a>`) would be the best option.
If, for some reason, you need to use a non-semantic element (like this `div`), you should make sure you provide all the necessary attributes to make it acessible (role, tabIndex, onKeyDown, ...)

my-code/whats-in/src/App.jsx
```diff
@@ -0,0 +1,90 @@
...
+import Moviepage from './pages/Movie'
+import Header from './components/Header'
+
+function App() {
+  const [searchText, setSearchText] = useState('')
+  const [movies, setMovies] = useState()
+  const [selectedMovie, setSelectedMovie] = useState(null)
+  const [selectedMovieInfo, setSelectedMovieInfo] = useState()
+  const [isLoading, setIsLoading] = useState(false)
+  const [likedMovies, setLikedMovies] = useState([]);
```
pdrbrnd, line 13: it would have been interesting to see some sort of approach to persist the favourites. currently they're gone after a refresh.

my-code/whats-in/src/App.jsx
```diff
@@ -0,0 +1,90 @@
...
+import React, { useEffect, useState } from 'react'
+import Homepage from './pages/Homepage'
+import Moviepage from './pages/Movie'
+import Header from './components/Header'
+
+function App() {
+  const [searchText, setSearchText] = useState('')
+  const [movies, setMovies] = useState()
+  const [selectedMovie, setSelectedMovie] = useState(null)
+  const [selectedMovieInfo, setSelectedMovieInfo] = useState()
```
pdrbrnd, line 11: you have an architectural problem here.
the source of truth for whether you have a movie selected + the selected movie info is this local state at `<App>`.
however, you do have a router that renders the movie detail page based on the URL (i.e.: if we visit `/movie`, we will get the movie details screen).

**If you try to navigate to a movie detail and refresh, your app will simply crash**

You had two options:

_A) assume that this is an app where the movie details are not shareable and every user should start at the search screen (closer to what you currently have)._
For this to work you would have to render the movie detail page **based on the data you have in the state**, but never based on the route.
So something like:
```jsx
<div className="container">
  {selectedMovieInfo ? <MoviePage data={selectedMovieInfo} {...whateverPropsItNeeds} /> : <Homepage />}
</div>
```
This way you're guaranteeing that the MoviePage component will only render when the required data is available.

_B) A more standard approach where you have a search screen at the root path "/" and a details screen at a child path "/:id"._
This means that selecting a movie would be simply **navigating to `/its-id`**.
The MovieDetail page would have to be self-contained and would include the fetch (and loading, and whatever needed) in itself. This way, a user could bookmark or share a URL for a specific movie.

my-code/whats-in/src/App.jsx
```diff
@@ -0,0 +1,90 @@
...
+  const [isLoading, setIsLoading] = useState(false)
+  const [likedMovies, setLikedMovies] = useState([]);
+  const [isEmptySearch, setIsEmptySearch] = useState(true)
```
pdrbrnd, line 14: creating a state for whether you have an empty state or not is very error prone and can easily lead to bugs. currently you have to manually set this boolean based on the fetch response + whatever is in the search at the time: besides being unnecessarily overly complex, a refactor within the fetch function could break this behaviour if someone forgets to keep this updated.
A simple solution is to infer this boolean based on the data we have and render it accordingly:
```jsx
if (searchText && movies.length === 0) return <NotFound />

return (
  // return non-empty app
)
```

my-code/whats-in/src/App.jsx
```diff
@@ -0,0 +1,90 @@
+import { Routes, Route, useNavigate } from 'react-router-dom';
+import React, { useEffect, useState } from 'react'
+import Homepage from './pages/Homepage'
+import Moviepage from './pages/Movie'
+import Header from './components/Header'
+
+function App() {
+  const [searchText, setSearchText] = useState('')
+  const [movies, setMovies] = useState()
```
pdrbrnd, line 9: you're initialising this state as undefined but arguably it should be an empty array. mutating the state type (in this case from `undefined` to `array` to `string`) is a bad idea and can lead to several bugs

my-code/whats-in/src/App.jsx
```diff
@@ -0,0 +1,90 @@
...
+    setSearchText(text)
+  }
+
+  useEffect(() => {
+    fetch(`https://www.omdbapi.com/?s=${searchText}&apikey=${process.env.REACT_APP_API_KEY}`)
+    .then(response => response.json())
+    .then(data => {
+      if (searchText === '') {
+        setIsEmptySearch(true)
+        setMovies('')
```
pdrbrnd, line 27: again, it's strange for `movies` to be a string.
I'm not sure, but probably `data.Search` is always an array, so why not just save it in local state as is?
```jsx
.then((data) => setMovies(data.Search))
```

my-code/whats-in/src/App.jsx
```diff
@@ -0,0 +1,90 @@
...
+    })
+  }, [searchText])
+
+  const handleMovieSelection = (movie) => {
+    setSelectedMovie(movie)
+  }
+
+  useEffect(() => {
+    if (selectedMovie && selectedMovie.imdbID) {
+      setIsLoading(true)
```
pdrbrnd, line 44: the name of this variable is a bit questionable:
given you have two separate network calls within this component, how do we know to which the loading refers to?
You could also have a loading indicator for when the user is searching something, so it would make sense to name this `loadingDetails` or similar

my-code/whats-in/src/App.jsx
```diff
@@ -0,0 +1,90 @@
...
+      fetch(`https://www.omdbapi.com/?i=${selectedMovie.imdbID}&apikey=${process.env.REACT_APP_API_KEY}`)
+        .then(response => response.json())
+        .then(data => {
+          setSelectedMovieInfo(data)
+          navigate('/movie')
+          setSelectedMovie(null)
+        })
```
pdrbrnd, line 51: again, an optimization could be made by simply inferring from the data we have.
it's unnecessary (and objectively worse because you have to be more careful and you're making it possible to have inconsistent data: what happens if you have `selectedMovie` but no `selectedMovieInfo`?) to have both this states. You could infer that you have a movie selected by checking if you have data in `selectedMovieInfo`.

Also, as stated above, you're navigating as a side-effect of loading movie data, where it should be:
A) no navigation and render components based on available data
B) exactly the opposite and simply navigate and fetch because of the route change

my-code/whats-in/src/App.jsx
```diff
@@ -0,0 +1,90 @@
...
+
+  const handleLike = (movie) => {
+    const movieFound = likedMovies.find((obj) => {return obj.imdbID === movie.imdbID})
+    
+    if (likedMovies.find((obj) => obj.imdbID === movie.imdbID)) {
+      setLikedMovies(likedMovies.filter((obj) => obj !== movieFound))
+    } else {
+      setLikedMovies((prevMovies) => [...prevMovies, movie])
+    }
+  }
```
pdrbrnd, line 69: you're defining a `movieFound` but then making the exact same operation below, when it could be `if (movieFound)`.

But a more paradigmatic approach for a function like this (adding to an array if it's not there yet), would be maybe something like:
```jsx
// assuming that the `likedMovies` is simply an array of IDs
// it's not necessary for us to save the entire movie when we only care for whether a particular movie is liked or not

const handleLike = (id) => {
  setLikedMovies((prev) => prev.includes(id) ? prev.filter((like) => like !== id) : [...prev, id])
}
```

my-code/whats-in/src/App.jsx
```diff
@@ -0,0 +1,90 @@
...
+
+  const isMovieLiked = (movie) => {
```
pdrbrnd, line 71: `likedMovies` should probably be an array of IDs. this state only cares whether a movie was liked or not so it just needs to save an identifier.

it would also mean that we wouldn't need this function, given we could simply check if `likedMovies.includes(movie.id)`

my-code/whats-in/src/pages/Homepage/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+import iconSearch from '../../images/icon-magnifier-grey.svg'
+import iconLike from '../../images/icon-heart-grey.svg'
+import iconLiked from '../../images/icon-heart-full.svg'
+import Loading from '../../components/Loading'
+import Empty from '../../components/Empty'
+import NotFound from '../../components/Notfound'
+
+export default function Homepage({ handleSearch, movies, handleMovieSelection, isLoading, handleLike, isMovieLiked, isEmptySearch }) {
+
+    const handleChange = (event) => {
```
pdrbrnd, line 11: the search network call should be debounced somehow (there are several valid approaches).

To understand why the current approach might be problematic:
1. Open up your browser's inspector and focus on the network tab
2. Focus your input and write something like "The godfather"
3. Check how many requests were fired to search for movies (one for each character input)

In this cases we want to wait for the user to stop typing _and then_ fire the search network call.

my-code/whats-in/src/pages/Homepage/index.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react'
+import iconSearch from '../../images/icon-magnifier-grey.svg'
+import iconLike from '../../images/icon-heart-grey.svg'
+import iconLiked from '../../images/icon-heart-full.svg'
+import Loading from '../../components/Loading'
+import Empty from '../../components/Empty'
+import NotFound from '../../components/Notfound'
+
+export default function Homepage({ handleSearch, movies, handleMovieSelection, isLoading, handleLike, isMovieLiked, isEmptySearch }) {
```
pdrbrnd, line 9: because of the architecture you chose of having all state in your App component, you're receiving a lot of props here when, in fact, this component could be independent with self-contained state.

Even if you still chose to go with option A with no routing, by inferring some logic based on state you already have you could simplify a bit the amount of props received.

my-code/whats-in/src/pages/Movie/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+import logoImdb from '../../images/logo-imdb.svg'
+import logoRottenTomatoes from '../../images/logo-rotten-tomatoes.svg'
+import iconLike from '../../images/icon-heart-grey.svg'
+import iconLikeWhite from '../../images/icon-heart-white.svg'
+
+export default function Moviepage( {movie, isMovieLiked, handleLike, likedMovies} ) {
+    const navigation = useNavigate();
+
+    const handleNavigation = () => {
+        navigation(-1)
```
pdrbrnd, line 13: nothing guarantees that the user came from the homepage. with this route-based approach, it could have opened directly the movie detail.

my-code/whats-in/src/pages/Movie/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    const checkRating = () => {
+        const rottenTomatoesRating = movie.Ratings.find((rating) => rating.Source === "Rotten Tomatoes");
+        return rottenTomatoesRating ? rottenTomatoesRating.Value : 'N/A'
+    }
+
+    const formatDetailsList = (info) => {
+        return info.split(', ').map((item) =>  <li>{item}</li>)
+    }
```
pdrbrnd, line 23: this is not formatting the detailsList but rather rendering it.
you could either:
- keep this `formatDetailsList` but rewrite it to just return an array of strings: `info.split(', ').map((item) => item)`
- rename this to `renderDetailsList` (not recommended)
- create a component like `<Details />` that would receive a string and render a list (recommended)
- just do it all inline.

also, it's good practice to add a `key` to an element when rendering lists
```jsx

```

my-code/whats-in/src/pages/Movie/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+                    <div className='movie-info__rating-bar__item'>
+                        <img className='movie-info__rating-bar__item__image--red' src={logoRottenTomatoes} alt="Imdb logo" />
+                        <p className='movie-info__rating-bar__item__text'>{checkRating()}</p>
+                    </div>
+                    <button onClick={() => handleClick(movie)} className={`movie-info__rating-bar__button ${isMovieLiked(movie) ? 'active' : ''}`}>
+                        <img src={isMovieLiked(movie) ? iconLikeWhite : iconLike} className='movie-info__rating-bar__button__icon' alt="Add to favorite button" />
+                        <img src={iconLikeWhite} className='movie-info__rating-bar__button__icon__white' alt="Add to favorite button" />
+                        <p className='movie-info__rating-bar__button__text'>{isMovieLiked(movie) ? 'Added' : 'Add to favourites'}</p>
+                    </button>
+                    {console.log(likedMovies)}
```
pdrbrnd, line 55: forgotten console log

pdrbrnd (commented):
Hey @fernanda-freitas thanks a lot for the submission :)
I'm very sorry for the delayed review but time has been scarce.

I hope my comments are helpful!

pdrbrnd:
@fernanda-freitas I went ahead and closed this PR simply for housekeeping reasons. Feel free to still respond to the comments and we'll do our best to further explain any doubts you might have.

my-code/whats-in/src/App.jsx
fernanda-freitas: Thank you Pedro! I also believe that amongst all, this is the most critical point and I just didn't know how to think of it at the time. I will definitely try to enhance the app based on your explanation.

